PR #11, opened 2019-03-29
nix/nixpkgs.nix: default `enableMozillaOverlay` to `false`.

Profpatsch:
We only need to enable the Overlay for rust nightly, which we only use
in our dev-shell for lorri. In all other places we don’t need it.

Having no default value also broke the tutorial (direnv setup),
because `nix-env -f ./nix/nixpkgs.nix` would not work anymore.


Profpatsch:
Merging because [multiple people have the same problem](https://github.com/target/lorri/pull/13).

PR #13, opened 2019-03-29
Default enableMozillaOverlay to false

DanBurton:
When following the tutorial's instructions for installing direnv from the overlay, I got this error:

```
$ nix-env -f ../nix/nixpkgs.nix -i -A direnv
error: cannot auto-call a function that has an argument without a default value ('enableMozillaOverlay')
```

I added a default value so that the command would work. If this is not preferred, I'd recommend updating the installation instructions.

### DIFF
--- a/nix/nixpkgs.nix
+++ b/nix/nixpkgs.nix
@@ -1,4 +1,4 @@
-{ enableMozillaOverlay }:
+{ enableMozillaOverlay ? false }:
 let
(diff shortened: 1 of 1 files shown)


DanBurton:
Dupe of #11

Profpatsch:
We noticed that, thanks. There is an open PR in https://github.com/target/lorri/pull/11. Was stalling it to set up CI first, but I think we can merge it right now.

PR #21, opened 2019-04-01
Document unexpected reevaluations

dudebout:


README.md
```diff
@@ -69,6 +69,15 @@ $ cd lorri
 $ nix-env -if .
 ```
 
+### (for non NixOS) Create gcroots directory
+
+lorri uses the `/nix/var/nix/gcroots/per-user/$USER` directory which
+is present in NixOS installation, but not always otherwise:
+
+```
+$ mkdir -m 0755 -p /nix/var/nix/gcroots/per-user/$USER
```
Profpatsch: This is fixed by https://github.com/target/lorri/pull/28 so we don’t need this bit of documentation anymore.

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
 ```
 
+### `lorri` reevaluates more than expected
+
+`lorri` sometimes recursively watches a directory that the user did
+not expect. This can happen for a number of reasons:
+
+1. when using a directory instead of a channel for `nixpkgs`, `lorri`
+   watches that directory recursively, including the potential `.git`
+   directory and its subdirectories
```
Profpatsch: `when using a local checkout instead`

I think the second part of the sentence is redundant, but you could say `, and will trigger on any file change`.

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+2. when specifying `src` via a path, `lorri` watches that path
```
Profpatsch: ```(like the much-used `src = ./.;`)```

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+   recursively (see https://github.com/target/lorri/issues/6
+   for details). To get around this, use a `filterSource`-based
```
Profpatsch: `builtins.filterSource`

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+   expression to specify `src`, e.g., use
```
Profpatsch: ```function to filter `src` ```

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+   [nix-gitignore](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/nix-gitignore/default.nix):
```
Profpatsch: This should be a permalink.

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+   `src = pkgs.nix-gitignore.gitignoreSource [] ./.` (note that the
```
Profpatsch: `or one of the functions in <link to nixpkgs/lib/sources.nix>`

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+   trailing `./.` will not be watched by `lorri`)
```
Profpatsch: I don’t understand the remark in parens.

README.md
```diff
@@ -101,6 +110,25 @@ Set these environment variables when debugging:
...
+3. when using a construct like `import ./.` to import a `default.nix`
+   file, `lorri` watches the current directory recursively. To get
+   around it, use `import ./default.nix`.
```
Profpatsch: Is that true? Normally `import ./.` is an alias for `./default.nix`, so I’m suprised this happens.

grahamc: Interesting, I'll need to explore this.

Profpatsch:
I applied my suggestions.

README.md
grahamc: I forgot to post my findings, but _yes_ this is true! Nix doesn't emit `/path/to/default.nix` as the true file, it emits just '/path/to`. A possible solution here is to append  a `/default.nix` when an imported directory is identified.

Profpatsch: Okay, opened a separate issue.

dudebout:
> Thanks for the docs! I have a few comments, if you want I can apply the suggestions.

Sorry @Profpatsch I missed this message. I am watching the repo and did not see it in all the recent updates. All your comments made sense. Regarding your question about the remark in parens. I was noting that despite the fact that we use the string `./.` we did not hit the issue where lorri watches the full directory.

README.md
curiousleo: I believe this is now fixed: https://github.com/target/lorri/blob/master/src/builder.rs#L160-L170

See https://github.com/target/lorri/pull/249

PR #24, opened 2019-04-03
direnv: print error if lorri watch has not been run

### LINKED ISSUE #9: Non-great error when entering a directory before running `lorri watch`
```
direnv: loading .envrc                                                                                                                                                                                                                
./.envrc: line 66: /home/michael/.cache/lorri/example/gc_root/build-0: No such file or directory
direnv: export +IN_NIX_SHELL
```

Seems like this should be detectable!

grahamc:
Closes #9, though we should make a new issue to have it auto-evaluate the first time.


grahamc:
made: #53, thanks!

PR #26, opened 2019-04-03
Set up Travis

grahamc:
We used to use CircleCI, but no longer. Let's get this repo up and running with Travis.

A few notes:
 - the .travis.yml is generated from the .travis.yml.nix and changing one without the other will cause the bulid to fail
 - Nix tests were using the host's nixpkgs in nix-shell cargo build scenarios, so I (secretly in the doc tests) set the NIX_PATH to the bogus nixpgks for our reader's sake
 - the timing window on the mpsc test was too short for Travis, so I widened it a bit
 - `echo -n` doesn't work inside a single-user nix-build happening inside a nix-shell on macOS, so replace with `printf` which should be more stable.


.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: cargo build & linters
+    nix: 2.2.1
+    os: linux
+    script: 'set -e
+
+      nix-shell --run ci_check
+
+      cat $(nix-build ./.travis.yml.nix --no-out-link) > .travis.yml
+
+      git diff -q ./.travis.yml
```
Profpatsch: What’s this do?

grahamc: It causes the build to fail if .travis.yml xor .travis.yml.nix are changed.

PR #28, opened 2019-04-03
Create nix per-user gcroot directory if missing

Profpatsch:
See commit messages. I also improved the error messages for permission errors.
It would be nice to have an integration test (which has to be sandboxed), I just tested it manually for now, but I think it’s fine.

### DIFF
--- a/src/build_loop.rs
+++ b/src/build_loop.rs
@@ -78,6 +78,10 @@ impl BuildLoop {
                 Ok(())
             };
             match go() {
+                // TODO: Make err use Display instead of Debug.
+                // Otherwise user errors (especially for IO errors)
+                // are pretty hard to debug. Might need to review
+                // whether we can handle some errors earlier than here.
                 Err(err) => panic!("{}: {:?}", "Builder failed", err),
                 Ok(()) => {}
             }
--- a/src/roots.rs
+++ b/src/roots.rs
@@ -25,8 +25,19 @@ impl Roots {
         path.push(name);
 
         debug!("Adding root from {:?} to {:?}", store_path, path,);
-        ignore_missing(std::fs::remove_file(&path))?;
-        symlink(&store_path, &path)?;
+        check_permission(
+            ignore_missing(std::fs::remove_file(&path)),
+            &format!("Can’t remove {}", path.display()),
+        )?;
+
+        check_permission(
+            symlink(&store_path, &path),
+            &format!(
+                "Can’t symlink {} to {}",
+                store_path.display(),
+                path.display()
+            ),
+        )?;
 
         // this is bad.
         let mut root = PathBuf::from("/nix/var/nix/gcroots/per-user");
@@ -40,8 +51,14 @@ impl Roots {
         root.push(format!("{}-{}", self.id, name));
 
         debug!("Connecting root from {:?} to {:?}", path, root,);
-        ignore_missing(std::fs::remove_file(&root))?;
-        symlink(&path, &root)?;
+        check_permission(
+            ignore_missing(std::fs::remove_file(&root)),
+            &format!("Can’t remove {}", root.display()),
+        )?;
+        check_permission(
+            symlink(&path, &root),
+            &format!("Can’t symlink {} to {}", path.display(), root.display()),
+        )?;
 
         Ok(path)
     }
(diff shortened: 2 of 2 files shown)


src/build_loop.rs
```diff
@@ -72,6 +72,7 @@ impl BuildLoop {
     pub fn forever(&mut self) {
         loop {
             let mut go = || -> Result<(), SingleBuildError> {
+                // TODO: real error handling here
```
grahamc: What kind of real error handling? Reading this comment makes me think you're referring to the "waiter exited" expect call. Can you explain a bit more (in the comment) about what this means?

grahamc:
Looks like this needs to be `clippy` and `fmt`'d @Profpatsch :)

src/build_loop.rs
Profpatsch: Moved the comment a few lines further down and expanded.

Profpatsch:
> handle_io_error, ignore_missing, and check_permission looks like they would be at home in a Haskell codebase

`ignore_missing` is your original function, slightly refactored. I can copy the code of course.

src/roots.rs
```diff
@@ -25,41 +25,99 @@ impl Roots {
...
 
-fn ignore_missing(err: Result<(), std::io::Error>) -> Result<(), std::io::Error> {
-    if let Err(e) = err {
-        match e.kind() {
-            std::io::ErrorKind::NotFound => Ok(()),
-            _ => Err(e),
+/// Check if a directory exists and is a directory.
+/// Try to create it if nothing’s there.
+/// Panic for everything else.
+fn ensure_directory_exists(dir: &PathBuf) {
```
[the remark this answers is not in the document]
Profpatsch: We have to panic anyway, we can of course bubble the errors further up and panic later. But does it make sense as long as we can’t handle the error in any way? It just adds more code.

Or is there some necessary cleanup I missed?

Profpatsch:
I inlined `handle_io_error`, that should make it more “rusty”.

src/roots.rs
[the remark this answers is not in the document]
grahamc: Yeah, I think we should raise the io error up out of the function. It is really unpleasant having panics, I try to only use them when a panic would reveal a bug in the code itself, or in the very outer-most level of the program (directly in `main()`)

src/build_loop.rs
```diff
@@ -78,6 +78,10 @@ impl BuildLoop {
                 Ok(())
             };
             match go() {
+                // TODO: Make err use Display instead of Debug.
+                // Otherwise user errors (especially for IO errors)
+                // are pretty hard to debug. Might need to review
+                // whether we can handle some errors earlier than here.
```
grahamc: :+1: nice :)

src/roots.rs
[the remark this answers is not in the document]
grahamc: I still think this should return a Result which is `expect`'d at the caller.

src/roots.rs
```diff
@@ -25,41 +25,91 @@ impl Roots {
...
+        // The user directory sometimes doesn’t exist,
+        // but we can create it (it’s root but `rwxrwxrwx`)
+        ensure_directory_exists(&root);
+
         root.push(format!("{}-{}", self.id, name));
 
         debug!("Connecting root from {:?} to {:?}", path, root,);
-        ignore_missing(std::fs::remove_file(&root))?;
-        symlink(&path, &root)?;
+        check_permission(
```
grahamc: I don't understand why we panic on permission denied, where currently we otherwise pass an `io` error on up anyway. What is the purpose of adding the extra handling of the EPERM?

Profpatsch: The problem is that the panic errors by `io` are extremely nondescriptive, so I’d rather fail fast and tell the user the reason. We have to fail anyway, and this is I/O code.

I mean we could split them out into their own enum and handle them on the top layer, but I don’t see anything we gain from that at this stage (can be done later if needed).

Profpatsch: Actually, let me rebase on #55 and push it via the new interface.

grahamc: You're right that IO errors are quite nondescript. I wrote up an example of a nicer, more rust-like way of implementing this: https://github.com/target/lorri/commit/04121896b6bebb44bdb06350ff792da85f57d8c5 Check it out?

PR #56, opened 2019-04-09
Implement an integration test

grahamc:
Adds an integration test for the basic workings of Lorri, without ever shelling out to `lorri`.

This PR is based on https://github.com/target/lorri/pull/55, so skip to "project: print..." as the first commit.

The final commit is the best one, where the integration test actually comes to exist.


tests/direnv/direnvtestcase.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Get an environment value with a borrowed str in the deepest Option.
+    /// Makes asserts nicer, like:
+    ///
+    ///    assert!(env.get_env("foo"), Some(Some("bar"));
+    ///
+    /// Whereas, with a standard HashMap.get you'd have to do:
+    ///
+    ///    assert!(env.get("foo"), Some(Some(String::new("bar"))));
+    ///
+    pub fn get_env<'a, 'b>(&'a self, key: &'b str) -> Option<Option<&'a str>> {
```
Profpatsch: This cries for a more semantic enum, like:

```
enum {
  NonExisting, // None()
  IdkWhatDoesThisMean, // Some(None())
  String(String) // Some(Some(String))
}
```

grahamc: :fire:

grahamc:
The feedback here is :+1:. I'll address it soon, but right now I'm being chainsed by the nix-build inside using coreutils. Digging in to that.

src/ops/mod.rs
```diff
@@ -18,7 +18,23 @@ pub struct ExitError {
 }
 
 /// Final result from a CLI operation
-pub type OpResult = Result<(), ExitError>;
+pub type OpResult = Result<Option<String>, ExitError>;
+
+/// Return an OpResult with a final message to print before exit 0
+/// Note, the final message is possibly intended to be consumed
+/// by automated tests.
```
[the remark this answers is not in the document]
grahamc: Possibly so, but since we only have one use case, I don't think it is a good time. We can extend from here once we have use cases.

grahamc:
This could use a re-review! Hopefully it passes on Travis, though I tested it locally many times.

grahamc:
Disregard the final commit, which is based on #28.

grahamc:
Ok, still iterating on this one. Some of the problems on Travis are a bit stickier than I expected.

src/logged-evaluation.nix
```diff
@@ -38,13 +36,13 @@ let
     name = "lorri-keep-env-hack-${drv.name}";
 
     origBuilder = drv.builder;
-    builder = builtins.storePath nixConfig.shell;
+    builder = "/bin/sh";
```
Profpatsch:
```
# We can do this because nix treats the string `"/bin/sh"` specifically
# and mounts the path w/ closure into its sandbox. It’s theoretically possible
# that a user compiles nix without that path, but we don’t support it.
```

grahamc: Nice. Can you add this as a suggestion, and I'll just click "Accept" and then merge? :) (also, `nix` -> `Nix`)

PR #62, opened 2019-04-23
Make 'enable direnv' step more obvious

joepie91:
I got stuck on direnv magically not seeming to do anything after setting everything up. It took me a while to realize that I missed the "enable direnv according to its instructions" step (assuming that the Nix expression would do that for me), as it was a relatively unnoticeable footnote.

This PR changes that, by making enabling it a clear and discrete step in the setup process.


Profpatsch:
Looks good thanks.

Merging, the CI error is because of an update of `cargo fmt` which adds something to the lock file.

PR #70, opened 2019-04-30
Explicitly depend on Bash

grahamc:
stdenv uses Bash not POSIX sh for setup:

    $ nix show-derivation $(which bash)
    [...snip...]
      "builder": "/nix/store/n9acaakxahkv1q3av11l93p7rgd4xqsf-bootstrap-tools/bin/bash",
      "args": [
        "-e",
        "/nix/store/9krlzvny65gdc8s7kpb6lkx8cd02c25b-default-builder.sh"
      ]

Using POSIX sh caused syntax errors which were not caught in CI for
some reason :(.


nix/runtime.nix
```diff
@@ -0,0 +1,37 @@
...
+    # Nix with sandboxing requires every path used at build time be
+    # explicitly declared. If we simply passed in the paths, they
+    # would be copied in as sources. Using builtins.storePath we're
+    # able to tell Nix that, no, in fact, treat these not as sources
+    # to copy, but instead of a regular store path.
+    #
+    # Include the explicit closure, too, otherwise we'll get mysterious
+    # "file not found" errors due to the glibc interpreter being
+    # missing.
+    {
```
grahamc: storePath only works if we're in a /nix/store store, not if we're in the `nix-build` fake store. 

This is very annoying.

Nix has some magic in its `<nix/config.nix>` but I'm not 100% sure that is the right thing to do.

Generally frustrated.

nix/runtime.nix
```diff
@@ -0,0 +1,64 @@
...
+      path = "\''${tools}/bin";
+      builder = "\''${tools}/bin/bash";
+      closure = [
+    $(sed -E 's/^(.*)$/    (dep \1)/' ${runtimeClosureInfo}/store-paths)
+      ];
+    }
+    EOF
+
+    mv ./scratch.nix $out
+  '';
```
grahamc: Example output of this:

```nix
# Nix with sandboxing requires every path used at build time be
# explicitly declared. If we simply passed in the paths, they
# would be copied in as sources. Using builtins.storePath we're
# able to tell Nix that, no, in fact, treat these not as sources
# to copy, but instead of a regular store path.
#
# Include the explicit closure, too, otherwise we'll get mysterious
# "file not found" errors due to the glibc interpreter being
# missing.
let
  # Magic inspired by Nix's config.nix:
  # https://github.com/NixOS/nix/blob/f9a2ea44867cd1dbb408bca4df0ced806137b7f7/corepkgs/config.nix.in#L23
  #
  # If the dependency is in the Nix store we're using, refer to
  # it as a literal store path. If it isn't, refer to it "normally".
  #
  # This makes sandboxing happy when in a nix-build, and the
  # evaluation happy when in a «cargo build».
  tools_host = /nix/store/7yp1kp9m2s5cjklxj4l5rr9gcizfbd06-lorri-runtime-tools;
  pp = x: builtins.trace x x;

  # Compare the stringified version of the tools_host Nix store path
  # to the evaluator's stringified Nix store path. Otherwise,
  # Nix will read the sources in to the /nix/store, and, well,
  # you can only copy the /nix/store in to the /nix/store so many
  # times before you run out of disk space.
  dep = if ("${toString (dirOf tools_host)}" == "${toString builtins.storeDir}")
    then (builtins.trace "using storePath"  builtins.storePath)
    else (builtins.trace "using toString" toString) # assume we have no sandboxing
    ;

  tools = dep tools_host;

in {
  path = "${tools}/bin";
  builder = "${tools}/bin/bash";
  closure = [
    (dep /nix/store/4apxd9wcaamlss78hkv1bpm8yn1crfhb-bash-4.4-p23-man)
    (dep /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27)
    (dep /nix/store/7yp1kp9m2s5cjklxj4l5rr9gcizfbd06-lorri-runtime-tools)
    (dep /nix/store/cinw572b38aln37glr0zb8lxwrgaffl4-bash-4.4-p23)
    (dep /nix/store/d9s1kq1bnwqgxwcvv4zrc36ysnxg8gv7-coreutils-8.30)
    (dep /nix/store/hql9ki8x230src24ljb9fad7rxxpzal0-attr-2.4.48)
    (dep /nix/store/s83xl21h4qiz5yxlhr0n2bm1dl574mhw-acl-2.2.53)
  ];
}
```

nix/runtime.nix
```diff
@@ -0,0 +1,64 @@
...
+      # Magic inspired by Nix's config.nix:
+      # https://github.com/NixOS/nix/blob/f9a2ea44867cd1dbb408bca4df0ced806137b7f7/corepkgs/config.nix.in#L23
+      #
+      # If the dependency is in the Nix store we're using, refer to
+      # it as a literal store path. If it isn't, refer to it "normally".
+      #
+      # This makes sandboxing happy when in a nix-build, and the
+      # evaluation happy when in a «cargo build».
+      tools_host = ${tools};
+      pp = x: builtins.trace x x;
```
Profpatsch: unused

nix/runtime.nix
```diff
@@ -0,0 +1,64 @@
...
+      # to the evaluator's stringified Nix store path. Otherwise,
+      # Nix will read the sources in to the /nix/store, and, well,
+      # you can only copy the /nix/store in to the /nix/store so many
+      # times before you run out of disk space.
+      dep = if ("\''${toString (dirOf tools_host)}" == "\''${toString builtins.storeDir}")
+        then (builtins.trace "using storePath"  builtins.storePath)
+        else (builtins.trace "using toString" toString) # assume we have no sandboxing
+        ;
+
+      tools = dep tools_host;
```
Profpatsch: Is this intended to be a circular definition? Further up you say `tools_host = ${tools}` (which could also be written as `tools_host = tools`.

Profpatsch: Hm, there’s a second `tools` further up. `let` being recursive, I’d have expected this to throw an error. For clarity, I would rename one of the two `tools` variables.

nix/runtime.nix
```diff
@@ -0,0 +1,64 @@
...
+
+    in {
+      path = "\''${tools}/bin";
+      builder = "\''${tools}/bin/bash";
+      closure = [
+    $(sed -E 's/^(.*)$/    (dep \1)/' ${runtimeClosureInfo}/store-paths)
```
Profpatsch: Instead of doing the giant heredoc, I’d just write a `${runtimeClosureInfo}/store-paths` to nix converter and import the resulting file from the nix expression (that you can write with `writeText`). That removes one layer of quoting.

nix/runtime.nix
```diff
@@ -0,0 +1,64 @@
...
+    let
+      # Magic inspired by Nix's config.nix:
+      # https://github.com/NixOS/nix/blob/f9a2ea44867cd1dbb408bca4df0ced806137b7f7/corepkgs/config.nix.in#L23
+      #
+      # If the dependency is in the Nix store we're using, refer to
+      # it as a literal store path. If it isn't, refer to it "normally".
+      #
+      # This makes sandboxing happy when in a nix-build, and the
+      # evaluation happy when in a «cargo build».
+      tools_host = ${tools};
```
Profpatsch: Should annotate that this expands to the nix store path at creation of this nix file.

shell.nix
```diff
@@ -42,7 +42,7 @@ pkgs.mkShell rec {
   # Set up a local directory to install binaries in
   CARGO_INSTALL_ROOT = "${LORRI_ROOT}/.cargo";
 
-  COREUTILS = "${pkgs.coreutils}";
+  CLOSURE_PARAMS = pkgs.callPackage ./nix/runtime.nix {};
```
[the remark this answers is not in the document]
Profpatsch: And get a comment for good measure.

src/logged-evaluation.nix
```diff
@@ -35,20 +37,20 @@ let
   keep-env-hack = drv: derivation (drv.drvAttrs // {
     name = "lorri-keep-env-hack-${drv.name}";
 
+    origExtraClosure = drv.extraClosure or [];
+    extraClosure = runtimeCfg.closure;
```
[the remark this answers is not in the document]
grahamc: I want to make sure we don't throw any data away, so each time I set a parameter we save it in to another variable (in the hope that `orig*` is unused. Maybe these should be named `lorriOrig...` instead for extra safety... but we definitely don't want to accidentally miss dependencies when setting up this drv. Make sense?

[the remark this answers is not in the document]
Profpatsch: Why not put all of the original attributes into one field?

PR #81, opened 2019-05-14
`lorri daemon`

Profpatsch:
Implements a `lorri daemon` command listening on a unix socket and a `lorri ping` command to send a build job to the daemon.

The test is still WIP, blocking on some action.

There’s still a few `TODO`s left, which will be addressed in a successive PR.


Profpatsch:
Okay, the test should now work.

nix/pre-check.sh
```diff
@@ -2,6 +2,9 @@ TEST_ROOT_DEST=${TMPDIR:-/tmp}/nix-test/nix
 mkdir -p $TEST_ROOT_DEST
 TEST_ROOT=$(realpath $TEST_ROOT_DEST)
 
+# prevent us from creating $HOME if it doesn’t exist
+export XDG_CACHE_DIR=${TMPDIR:-/tmp}/nix-test/xdg-cache-dir
```
grahamc: `export XDG_CACHE_HOME=...`

Profpatsch:
Heck! It’s green!

Profpatsch:
Btw, I’m not sure whether we really need the dependency on `nix` just to get access to the `poll` function. It’s a very, very thin wrapper around https://docs.rs/crate/libc/0.2.55, so might be best to just copy that one function over.

Profpatsch:
I want to get this PR merged, it’s been open for too long.

Since I documented every TODO in the comments, we can fix them in separate PRs.

grahamc:
Getting the existing TODOs todone won't be a substantial lift, and when possible, it is always better to get it done before integration. Let's do it in this PR.

shlevy:
Anything we're waiting on here?

src/daemon.rs
```diff
@@ -0,0 +1,114 @@
...
+        match ping {
+            Err(ReadError::Timeout) => debug!(
+                "Client didn’t send a `Ping` message after waiting for {}",
+                &self.read_timeout
+            ),
+            Err(ReadError::Deserialize(e)) => {
+                debug!("Client `Ping` message could not be decoded: {}", e)
+            }
+            Ok(p) => {
+                eprintln!("pinged with {}", p.nix_file);
```
grahamc: Should this be an `info!` message?

src/lib.rs
```diff
@@ -42,7 +43,40 @@ pub mod ops;
...
 pub mod watch;
 
+use std::path::PathBuf;
+
 // OUT_DIR and build_rev.rs are generated by cargo, see ../build.rs
 include!(concat!(env!("OUT_DIR"), "/build_rev.rs"));
+
+/// A .nix file.
+#[derive(Hash, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
+pub struct NixFile(PathBuf);
```
grahamc: :ok_hand:

src/ops/direnv/mod.rs
```diff
@@ -15,8 +17,24 @@ pub fn main(project: &Project) -> OpResult {
...
+            .unwrap();
+    } else {
+        eprintln!("Uh oh, your lorri daemon is not running.");
+    }
+
+    // TODO: this needs to move to a daemon message
     if !shell_root.exists() {
-        return ExitError::errmsg("Please run 'lorri watch' before using direnv integration.");
+        return Err(ExitError::errmsg(
+            "Please run 'lorri watch' before using direnv integration.",
```
grahamc:
```suggestion
            "Please start `lorri daemon` or run `lorri watch` before using direnv integration.",
```

src/ops/direnv/mod.rs
```diff
@@ -15,8 +17,24 @@ pub fn main(project: &Project) -> OpResult {
     let mut shell_root = project.gc_root_path().unwrap();
     shell_root.push("build-0"); // !!!
 
+    // TODO: don’t start build/evaluation automatically, let the user decide
```
grahamc: I don't understand the TODO, a ping should be always desirable, no?

src/ops/direnv/mod.rs
```diff
@@ -15,8 +17,24 @@ pub fn main(project: &Project) -> OpResult {
...
+            .write(&Ping {
+                nix_file: project.expression().clone(),
+            })
+            .unwrap();
+    } else {
+        eprintln!("Uh oh, your lorri daemon is not running.");
+    }
+
+    // TODO: this needs to move to a daemon message
     if !shell_root.exists() {
```
grahamc: I wonder when under the circumstance:

 - if the daemon is up
 - and we ping successfully
 - and the shell root doesn't exist

if we should sleep until the initial evaluation is done?

Hmm maybe not, that might block editors... but perhaps the error below should be contextual, where in this situation we print a message saying it is not _yet_ ready.

src/constants.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Path constants like the GC root directory.
+pub struct Paths {
+    gc_root_dir: PathBuf,
+    daemon_socket_file: PathBuf,
+}
+
+impl Paths {
+    /// Set up project paths
+    pub fn new() -> std::io::Result<Paths> {
```
grahamc: Rust's coding style makes `new` never return errors. Maybe `get` or `build` or `initialize` is a better choice here?

src/ops/mod.rs
```diff
@@ -49,16 +57,16 @@ impl ExitError {
     /// Helpers to create exit results
     ///
     /// Note: err panics if exitcode is zero.
-    fn err<T>(exitcode: i32, message: T) -> OpResult
+    fn err<T>(exitcode: i32, message: T) -> ExitError
```
grahamc: (note for future graham: why are we changing OpResult to ExitError?)

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::socket::communicate::client;
+use crate::socket::communicate::{Ping, DEFAULT_READ_TIMEOUT};
+
+/// See the documentation for lorri::cli::Command::Shell for more
+/// details.
+pub fn main(nix_file: NixFile) -> OpResult {
+    // TODO: set up socket path, make it settable by the user
+    client::ping(DEFAULT_READ_TIMEOUT)
+        // TODO
+        .connect(&::socket::path::SocketPath::from(
```
grahamc: I think we might want to abstract out this SocketPath::from stuff in to a function we can call

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+            ::ops::get_paths()?.daemon_socket_file(),
+        ))
+        .unwrap()
+        .write(&Ping { nix_file })
+        .unwrap();
+    ok()
+}
+
+// /// Handle the ping
```
grahamc:
```suggestion
```

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+// // the ReadWriter here has to be the inverse of the `Client.ping()`, which is `ReadWriter<!, Ping>`
```
grahamc:
```suggestion
```

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+// fn ping(rw: ReadWriter<Ping, NoMessage>) {
```
grahamc:
```suggestion
```

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+//     // tx.send(rw.read())
```
grahamc:
```suggestion
```

src/ops/ping.rs
```diff
@@ -0,0 +1,28 @@
...
+// }
```
grahamc:
```suggestion
```

src/project.rs
```diff
@@ -1,23 +1,19 @@
...
 use std::io;
 use std::os::unix::ffi::OsStrExt;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use NixFile;
 
 /// A specific project which we are operating on
 #[derive(Debug)]
-pub struct Project {
+pub struct Project<'a, 'b> {
```
grahamc: :ok_hand:

src/project.rs
```diff
@@ -31,68 +27,28 @@ pub enum ProjectLoadError {
...
+    pub fn expression(&self) -> &NixFile {
+        &self.nix_file
     }
 
     /// Absolute path to the projects' gc root directory, for pinning
     /// build and evaluation products
     pub fn gc_root_path(&self) -> Result<PathBuf, std::io::Error> {
-        let path = self.gc_root.join(self.name()).join("gc_root");
+        // TODO: use a hash of the project’s abolute path here
+        // to avoid collisions
```
grahamc: is this a thing we're doing now?

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// We declare 1s as the time readers should wait
+/// for the other side to send something.
+pub const DEFAULT_READ_TIMEOUT: Timeout = Timeout::from_millis(1000);
+
+/// Enum of all communication modes the lorri daemon supports.
+#[derive(Serialize, Deserialize)]
+pub enum CommunicationType {
+    /// Ping the daemon from a project to tell it to watch & evaluate
+    // TODO: rename to BuildProject or something?
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    Ping,
+    // /// Listen for events on the daemon
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    // DaemonListener,
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    // /// Get the daemon status
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    // DaemonStatus,
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+/// watching `nix_file`. See `CommunicationType::Ping`.
+#[derive(Serialize, Deserialize)]
+pub struct Ping {
+    /// The nix file to watch and build on changes.
+    pub nix_file: NixFile,
+}
+
+/// No message can be sent through this socket end (empty type).
+pub enum NoMessage {}
+// pub struct DaemonEvent {}
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+// pub struct DaemonStatus {}
```
grahamc:
```suggestion
```

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    #[derive(Debug, Serialize, Deserialize)]
+    pub struct ConnectionAccepted();
+
+    /// Server-side part of a socket transmission,
+    /// listening for incoming messages.
+    pub struct Listener {
+        /// Bound Unix socket.
+        listener: UnixListener,
+        /// Lock that keeps our socket exclusive.
+        #[allow(dead_code)]
```
grahamc: What is this `allow` about?

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// A `Client` that can talk to a `Listener`.
+    pub struct Client<R, W> {
+        /// Type of interaction with the `Listener`.
+        comm_type: CommunicationType,
+        /// Connected socket.
+        socket: Option<UnixStream>,
+        /// Timeout for reads/writes.
+        timeout: Timeout,
+        phantom_r: PhantomData<R>,
+        phantom_w: PhantomData<W>,
```
grahamc: Maybe should change these `phantom_*` names to indicate why they're there, not that they're phantom. Like to hint that they are used for pinning the read / write data types.

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+                comm_type,
+                socket: None,
+                timeout,
+                phantom_r: PhantomData,
+                phantom_w: PhantomData,
+            }
+        }
+
+        /// Connect to the `Listener` listening on `socket_path`.
+        pub fn connect(self, socket_path: &SocketPath) -> Result<Client<R, W>, InitError> {
```
grahamc: Hmm maybe `connect` should return a ConnectedClient or something so that `socket` ins't an Option. Not sure. What do you think?

src/socket/communicate.rs
```diff
@@ -0,0 +1,239 @@
...
+        }
+    }
+
+    /// Client for the `Ping` communication type.
+    /// Reading and writing messages is bounded by `timeout`.
+    pub fn ping(timeout: Timeout) -> Client<NoMessage, Ping> {
+        Client::bake(timeout, CommunicationType::Ping)
+    }
+
+    // /// Client for the `DaemonListener` communication type
```
grahamc: Let's delete this. (I'd do it, but this is a lot of lines to add single-line suggestions.)

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+pub struct ReadWriter<'a, R, W> {
+    // where R: serde::Deserialize {
+    socket: &'a UnixStream,
+    phantom_r: PhantomData<R>,
+    phantom_w: PhantomData<W>,
+}
+
+/// Milliseconds accepted by a `Timeout`.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct Millis(u16);
```
grahamc: Hmm why make our own when Duration can work okay too?

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub const fn from_millis(m: u16) -> Timeout {
+        Timeout::D(Millis(m))
+    }
+}
+
+impl std::fmt::Display for Timeout {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Timeout::Infinite => write!(f, "infinite timeout"),
+            Timeout::D(Millis(ms)) => write!(f, "{}ms", ms),
```
grahamc: let's make `infinite timeout` `infinity`, or change `{}ms` to `{}ms timeout`

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+
+// put into the Io part of a `bincode::ErrorKind`
+fn into_bincode_io_error<T>(res: std::io::Result<T>) -> bincode::Result<T> {
+    res.map_err(|e| Box::new(bincode::ErrorKind::Io(e)))
+}
+
+/// Run action with a timeout, return the remaining timeout
+/// or `Err` if there’s no time remaining.
+///
+/// The interface is not very “rusty”, but it’s only used internally.
```
grahamc: I don't love it, but okay ;)

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    /// Wait for a message to arrive.
+    pub fn read(&self, timeout: &Timeout) -> Result<R, ReadError>
+    where
+        R: serde::de::DeserializeOwned,
+    {
+        let timeout_socket = timeout::TimeoutReadWriter::new(self.socket, timeout);
+
+        // XXX: “If this returns an Error, `reader` may be in an invalid state”.
```
grahamc: I think that means the socket is tainted and should be restarted, since a read may resume in a weird spot in the data stream.

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    use self::nix::libc;
+    use self::nix::poll;
+    use super::{Millis, Timeout};
+    use std::os::unix::io::AsRawFd;
+    use std::os::unix::net::UnixStream;
+
+    /// Wait until `to_fd` receives the poll event from `events`, up to `timeout` length
+    /// of time.
+    /// Copied from https://docs.rs/crate/timeout-readwrite/0.2.0/source/src/utils.rs
+    /// written by Jonathan Creekmore and published under Apache-2.0.
```
grahamc: mmm maybe we should use that crate?

src/socket/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+        })?;
+
+        into_bincode_io_error(self.socket.flush())?;
+
+        Ok(())
+    }
+}
+
+/// Wrap a socket with a timeout. Inspired by https://docs.rs/crate/timeout-readwrite/0.2.0/.
+mod timeout {
```
grahamc: I don't love the code in this module, it feels quite complex and I don't really want to "own" it.

src/main.rs
```diff
@@ -3,70 +3,85 @@ extern crate structopt;
...
     lorri::logging::init_with_default_log_level(opts.verbosity);
     debug!("Input options: {:?}", opts);
 
-    let project = Project::from_cwd();
-
-    let result: OpResult = match (opts.command, project) {
-        (Command::Info, Ok(project)) => info::main(&project),
-
-        (Command::Build, Ok(project)) => build::main(&project),
+    match lorri::ops::get_paths() {
```
grahamc: Let's undo this part. Many commands present and future may not require a local project in the current directory, which is something the previous version of this supported explicitly. This will fix my first issue with the daemon:

```
[grahamc@Petunia:~]$ lorri daemon
There is no `shell.nix` in the current directory (/home/grahamc)
You can use the following minimal `shell.nix` to get started:

let
  pkgs = import <nixpkgs> {};
in
  pkgs.mkShell {
    buildInputs = [
        pkgs.hello
    ];
  }
```

grahamc: We talked about this: having the shell.nix check explicitly means we can give a nicer more specific error message, so +1 on what is here now.

Profpatsch:
I had to rebase, sorry, otherwise the `release.nix` would conflict (and I couldn’t generate the release number from a merge commit).

First unreviewed commit: 07eac0d, “main.rs: `daemon` subcommand doesn’t require a shell.nix in cwd”

src/ops/direnv/mod.rs
grahamc: Yeah, this is definitely needed. The UX currently is quite confusing:

```
[grahamc@Petunia:~/projects/github.com]$ cd NixOS/ofborg
direnv: loading .envrc
Uh oh, your lorri daemon is not running.
Please run 'lorri watch' before using direnv integration.

[grahamc@Petunia:~/projects/github.com/NixOS/ofborg]$ cd ..
direnv: unloading

   <--- here I go start `lorri daemon` elsewhere -->

[grahamc@Petunia:~/projects/github.com/NixOS]$ cd ofborg
direnv: loading .envrc
Please run 'lorri watch' before using direnv integration.

    <!-- lorri daemon starts processing -->

[grahamc@Petunia:~/projects/github.com/NixOS/ofborg]$ cd ..
```

src/cli.rs
```diff
@@ -42,6 +43,14 @@ pub enum Command {
     #[structopt(name = "watch")]
     Watch,
 
+    /// TODO
+    #[structopt(name = "daemon")]
+    Daemon,
+
+    /// TODO remove
```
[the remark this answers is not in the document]
Profpatsch: Should we prefix plumbing commands by `_`? `lorri _ping`

src/cli.rs
```diff
@@ -51,6 +60,16 @@ pub enum Command {
     Init,
 }
 
+// TODO remove
+#[derive(StructOpt, Debug)]
+/// Ping the daemon to start a build
+pub struct Ping {
+    // TODO
+    #[structopt(parse(from_os_str))]
+    /// The nix file to watch and build on changes.
```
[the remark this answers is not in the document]
Profpatsch: I left it in to satisfy the doc warning

src/daemon.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::project::Project;
+use crate::roots::Roots;
+use crate::socket::communicate::{NoMessage, Ping, DEFAULT_READ_TIMEOUT};
+use crate::socket::{ReadError, ReadWriter, Timeout};
+use crate::NixFile;
+use std::collections::HashMap;
+use std::sync::mpsc;
+
+/// Instructs the daemon to start a build.
+pub struct StartBuild {
```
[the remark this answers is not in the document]
Profpatsch: I’d love some other word than `ping` here, because a ping usually does not cause the server to behave differently.

src/daemon.rs
```diff
@@ -0,0 +1,114 @@
...
+        let tx = self.build_events_tx.clone();
+        let root_dir = self.paths.gc_root_dir().to_owned();
+
+        self.handler_threads
+            .entry(nix_file.clone())
+            .or_insert_with(|| {
+                // We construct a Project here for each dependency we get.
+                let project = Project::new(&nix_file, &root_dir);
+                // TODO unwrap
+                let roots = Roots::from_project(&project).unwrap();
```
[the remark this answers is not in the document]
Profpatsch: I think a bit of refactoring will make this work for all cases. The problem here is that it implicitly tries to create the root directory if it doesn’t exist, which can fail.

src/daemon.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Accept handler for `socket::communicate::Ping` messages.
+    /// For a valid ping message, it sends an instruction to start
+    /// the build to `build_chan`.
+    // TODO: make private again
+    // the ReadWriter here has to be the inverse of the `Client.ping()`, which is `ReadWriter<!, Ping>`
+    pub fn ping(&self, rw: ReadWriter<Ping, NoMessage>, build_chan: mpsc::Sender<StartBuild>) {
+        let ping: Result<Ping, ReadError> = rw.read(&self.read_timeout);
+        match ping {
+            Err(ReadError::Timeout) => debug!(
+                "Client didn’t send a `Ping` message after waiting for {}",
```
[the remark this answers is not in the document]
Profpatsch: Yeah, I implemented `Display` for `Timeout` that way.

src/ops/direnv/mod.rs
Profpatsch: No, most of the time I’m developing I don’t want to start any kind of build in the background, I just want it to put me in the environment of the last build. Though it’s course that’s debatably the whole point of lorri that the developer doesn’t have to think about it.

src/ops/direnv/mod.rs
Profpatsch: I never want to block the user by default, rather we should implement a command to wait for the next successful evaluation.

src/ops/ping.rs
Profpatsch: Hm, it’s a bit self-similar in places, but I don’t think it’ll stick around for too long.

src/project.rs
Profpatsch: Nope, not yet. Haven’t got around to implementing that.

src/socket/communicate.rs
Profpatsch: See above, I think we need a different name for `Ping`.

src/socket/communicate.rs
Profpatsch: Should we keep that around as a template for implementing more complex server messages?

src/socket/communicate.rs
Profpatsch: `bind_lock` is never explicitely used anywhere, the lock is dropped when `Listener` goes out of scope, so we can hide the warning.

src/socket/mod.rs
Profpatsch: Originally it was like that, but then I found out that the `poll` syscall can only accept timeouts of up to 16 bit ms length. So instead of making every single read/write return a `Result`, rather make it impossible to construct a timeout that is out of bounds.

src/socket/mod.rs
Profpatsch: Yeah, it’s not great. But it fixes a couple of bugs that were in the original logic & deduplicates that logic between `react` and `communicate`.

src/socket/mod.rs
Profpatsch: So in that case we’d have to deallocate the unix socket and create a new one?

src/socket/mod.rs
Profpatsch: Nah, it has a quite horrible interface and we just need this one bit of logic, so I’d rather copy that with a link back.

src/socket/mod.rs
Profpatsch: It’s sadly the only thing to not break the Darwin socket implementation in non-debuggable ways. Just setting the socket timeout option worked fine on Linux, but not on Darwin. It’s just a small wrapper around the `poll(2)` syscall.

src/constants.rs
```diff
@@ -0,0 +1,42 @@
...
+        let pd = ProjectDirs::from("com.github.target.lorri", "lorri", "lorri")
+            .expect("Could not determine lorri project/cache directories, please set $HOME");
+        let create_dir = |base: &Path, dir: &str| -> std::io::Result<PathBuf> {
+            let d = base.join(dir);
+            std::fs::create_dir_all(&d).and(Ok(d))
+        };
+        Ok(Paths {
+            gc_root_dir: create_dir(pd.cache_dir(), "gc_roots")?,
+            daemon_socket_file: pd.cache_dir().join("daemon.socket"),
```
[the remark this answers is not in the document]
Profpatsch: It should only fail if `HOME` does not exist at all (and the XDG variables don’t as well), so the `expect` message should be fine.

src/daemon.rs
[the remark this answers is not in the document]
Profpatsch: I’d like to move that to a different PR.

src/ops/direnv/mod.rs
Profpatsch: Can we move this to a different PR?

src/ops/direnv/mod.rs
Profpatsch: Also probably best to discuss in an issue and then decide there.

src/ops/mod.rs
Profpatsch: Returning an `OpResult` didn’t work in all cases I need `err` for, because I needed to construct an `ExitError` in some cases.

src/socket/communicate.rs
Profpatsch: Yeah, that would be a more type-safe way to do it. I just couldn’t be bothered in this case, because the module is pretty small and doesn’t expose any way for the user to get this wrong (`bake` is not exposed and `connect` already fills in the socket).

src/daemon.rs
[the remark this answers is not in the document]
grahamc: OK

src/ops/direnv/mod.rs
grahamc: Ok!

grahamc: Actually, please just accept my suggested change as-is since it is an improvement at least. The rest can be another PR

src/socket/communicate.rs
grahamc: Perhaps in an another PR?

src/socket/communicate.rs
grahamc: I don't think so, personally. I'd rather avoid keeping comments around

src/socket/mod.rs
grahamc: I think so

src/daemon.rs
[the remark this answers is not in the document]
Profpatsch: Okay, we talked about this and renamed it to `IndicateActivity` with an explanatory comment.

src/ops/direnv/mod.rs
Profpatsch: Done, issue at https://github.com/target/lorri/issues/99

src/ops/direnv/mod.rs
Profpatsch: issue is at: https://github.com/target/lorri/issues/100

src/ops/direnv/mod.rs
Profpatsch: Discussion in https://github.com/target/lorri/issues/99

src/socket/communicate.rs
Profpatsch: https://github.com/target/lorri/issues/101

src/socket/mod.rs
Profpatsch: https://github.com/target/lorri/issues/102

Profpatsch:
I *think* that is everything. Please review the final commits, then I can autosquash and we can merge.

PR #82, opened 2019-05-16
Prettier user-facing success message for self-upgrade

### LINKED ISSUE #80: pretty-print status in self-update
Running lorri self-update I get the following output:

```
Building ...
replacing old 'lorri'
installing 'lorri'
Ok(ExitStatus(ExitStatus(0)))
```

I guess the last part should be pretty-printed to something else (:nail_care: ).

gilligan:
Fixes #80

### DIFF
--- a/src/ops/upgrade/mod.rs
+++ b/src/ops/upgrade/mod.rs
@@ -64,11 +64,17 @@ pub fn main(upgrade_target: cli::UpgradeTo) -> OpResult {
     let tmpdir = tempdir().unwrap();
     match expr.clone().attribute("package").path(&tmpdir.path()) {
         Ok(build_result) => {
-            let ret = Command::new("nix-env")
+            let status = Command::new("nix-env")
                 .arg("--install")
                 .arg(build_result)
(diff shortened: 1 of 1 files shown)


src/ops/upgrade/mod.rs
```diff
@@ -64,11 +64,18 @@ pub fn main(upgrade_target: cli::UpgradeTo) -> OpResult {
...
     match expr.clone().attribute("package").path(&tmpdir.path()) {
         Ok(build_result) => {
-            let ret = Command::new("nix-env")
+            let status = Command::new("nix-env")
                 .arg("--install")
                 .arg(build_result)
-                .status();
-            Ok(Some(format!("{:?}", ret)))
+                .status() 
+                .expect("Error: failed to execute nix-env --install");
```
gilligan: The `expect` here means lorri will blow up in the users face when `nix-env` fails to execute but I suppose in that case exploding is probably justified.

Profpatsch:
Thanks!

PR #83, opened 2019-05-17
Implement "init" command

### LINKED ISSUE #41: `lorri init`
Set up a project to work with lorri and direnv.
 
For now that means:

* Make sure `shell.nix` exists and is a valid shell file
* `.envrc`

If the files don’t exist, we simply create them and populate them with our setup code. If they exist, we don’t touch them and print an informational message, like:

* “`shell.nix` exists, skipping. Make sure it is of a form that works with `nix-shell`.”
* “`.envrc` exists, skipping. Please add eval `"$(lorri direnv)"` to it to set up lorri support.”

In the case of `shell.nix` we could also try to find out whether it is a correct shell file.

gilligan:
Implements an `init` command that writes a default shell.nix and .envrc to $CWD if those files don't exist already.

Should fix #41 (?) however it does nothing about validating whether or not the existing `shell.nix` is valid - which is something @Profpatsch mentioned in #41

### DIFF
--- a/src/ops/init/mod.rs
+++ b/src/ops/init/mod.rs
@@ -1,6 +1,6 @@
 //! Bootstrap a new lorri project
 
-use crate::ops::{ok, ExitError, OpResult};
+use crate::ops::{ok, ok_msg, ExitError, OpResult};
 use std::fs::File;
 use std::io;
 use std::io::Write;
@@ -33,5 +33,5 @@ pub fn main(default_shell: &str, default_envrc: &str) -> OpResult {
     to_op(create_if_missing(Path::new("./shell.nix"), default_shell))?;
     to_op(create_if_missing(Path::new("./.envrc"), default_envrc))?;
 
-    Ok(Some(String::from("\nSetup done.")))
+    ok_msg(String::from("\nSetup done."))
 }
--- a/src/ops/init/mod.rs
+++ b/src/ops/init/mod.rs
@@ -6,9 +6,9 @@ use std::io;
 use std::io::Write;
 use std::path::Path;
 
-fn create_if_missing(path: &Path, contents: &str) -> Result<(), io::Error> {
+fn create_if_missing(path: &Path, contents: &str, msg: &str) -> Result<(), io::Error> {
     if path.exists() {
-        println!("- Skipping existing {}", path.display());
+        println!("- {} {}", msg, path.display());
         Ok(())
     } else {
         let mut f = File::create(path)?;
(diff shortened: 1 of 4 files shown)


src/ops/init/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+/// See the documentation for lorri::cli::Command::Init for
+/// more details
+pub fn main(default_shell: &str, default_envrc: &str) -> OpResult {
+    println!("Bootstrapping a new lorri project\n");
+
+    to_op(create_if_missing(Path::new("./shell.nix"), default_shell))?;
+    to_op(create_if_missing(Path::new("./.envrc"), default_envrc))?;
```
Profpatsch: I’d add the better “exists” messages I listed in #41 here:

* “shell.nix exists, skipping. Make sure it is of a form that works with nix-shell.”
* “.envrc exists, skipping. Please add eval "$(lorri direnv)" to it to set up lorri support.”

src/ops/init/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    Ok(Some(String::from("\nSetup done.")))
```
Profpatsch: nit: `ops::ok_msg()`

src/ops/init/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    match e {
+        Ok(_) => ok(),
+        Err(e) => ExitError::errmsg(format!("{}", e)),
+    }
+}
+
+/// See the documentation for lorri::cli::Command::Init for
+/// more details
+pub fn main(default_shell: &str, default_envrc: &str) -> OpResult {
+    println!("Bootstrapping a new lorri project\n");
```
Profpatsch: I think we can leave out this message, because the user ostensibly requests the bootstrap when she types `lorri init`.

Profpatsch: Or, or leave it in, but use it to print the directory: `"… in /home/user/current/working/dir"`.

src/ops/init/mod.rs
gilligan: Okay cool, will do

src/ops/init/mod.rs
gilligan: Actually I think you were right - I agree that I should just remove it

src/ops/init/mod.rs
gilligan: A bit wordy for my taste but on the other hand i don't have strong feelings about that ;-)

PR #87, opened 2019-05-31
Travis cargo cache

Profpatsch:
Add caching to the travis (non-nix-build) jobs.

Before: 
screenshot

After:
screenshot

Now both runs spend most of their time in fetching the nightly rust toolchain and the MacOS builder spends another 2 minutes or so setting up a multi-user nix daemon when a single-user install would suffice.

The `nix-build` runs could also be sped up by passing the cargo cache manually to the build.

### DIFF
--- a/.travis.yml.nix
+++ b/.travis.yml.nix
@@ -55,6 +55,7 @@ let
         in (map rmTarget [
           "lib${projectname}.rlib"
           # our own binaries/libraries (keep all other deps)
+          "${projectname}*"
           "build/${projectname}-*"
           "deps/${projectname}-*"
           "deps/lib${projectname}-*"
@@ -62,6 +63,10 @@ let
           ".fingerprint/${projectname}-*"
           # build script executable
           "incremental/build_script_build-*"
+          # TODO: the direnv integration test is not deterministic
+          "direnv-*"
+          "deps/direnv-*"
+          "incremental/direnv-*"
         ]);
         # TODO: this might improve things, but we don’t want
         # to open another `nix-shell` (because it takes a few seconds)
--- a/.travis.yml.nix
+++ b/.travis.yml.nix
@@ -42,6 +42,8 @@ let
         travis_fold travis-yml-idempotent \
           git diff -q ./.travis.yml
       '';
+      before_cache = [ "rm -rf /home/travis/.cargo/registry" ];
+      cache.directories = [ "/home/travis/cargo" ];
     };
   };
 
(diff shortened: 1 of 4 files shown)


.travis.yml.nix
```diff
@@ -35,13 +37,42 @@ let
...
+          "deps/${projectname}-*"
+          "deps/lib${projectname}-*"
+          "incremental/${projectname}-*"
+          ".fingerprint/${projectname}-*"
+          # build script executable
+          "incremental/build_script_build-*"
+          # TODO: the direnv integration test is not deterministic
+          "direnv-*"
+          "deps/direnv-*"
+          "incremental/direnv-*"
```
grahamc: Will it be a problem when this list goes stale?

.travis.yml.nix
```diff
@@ -35,13 +37,42 @@ let
...
+    };
+
+    # cache rust dependency building
+    cache = name: {
+      before_cache =
+        let rmTarget = path: "rm -rvf $TRAVIS_BUILD_DIR/target/debug/${path}";
+        in (map rmTarget [
+          "lib${projectname}.rlib"
+          # our own binaries/libraries (keep all other deps)
+          "lorri*"
```
grahamc: why `lorri` here instead of `${projectname}`?

.travis.yml.nix
```diff
@@ -35,13 +37,42 @@ let
...
+          "build/${projectname}-*"
```
grahamc: why the abstraction instead of just typing `lorri`?

.travis.yml.nix
```diff
@@ -35,13 +37,42 @@ let
...
           git diff -q ./.travis.yml
       '';
+      # delete all our own artifacts from the cache dir
+      # based on https://gist.github.com/jkcclemens/000456ca646bd502cac0dbddcb8fa307
+    };
+
+    # cache rust dependency building
+    cache = name: {
+      before_cache =
+        let rmTarget = path: "rm -rvf $TRAVIS_BUILD_DIR/target/debug/${path}";
```
grahamc:
```suggestion
        let rmTarget = path: "rm -rvf \"$TRAVIS_BUILD_DIR/target/debug/${path}\"";
```

grahamc:
Looks nice :)

.travis.yml.nix
Profpatsch: It will cache more things than it needs to. But unless `cargo` changes its `target/` directory structure, I doubt much will change here.

.travis.yml.nix
Profpatsch: Makes it easier to see the structure I think.

Profpatsch:
Rebased on master

PR #93, opened 2019-06-07
Actually support Go

grahamc:
#75 implemented support for shellHooks, but ... oops: Go doesn't use shellHooks to setup GOPATH.

This code monkeypatches much sooner: https://github.com/NixOS/nixpkgs/blob/22a765d0da7070311e5b8c2bb29fa496c833dae5/pkgs/stdenv/generic/setup.sh#L238-L284

and correctly supports GOPATH.

It also fixes a bit of an annoying bug where an appended variable would be appended, including a delimiter, even if the variable was empty before: https://github.com/target/lorri/commit/664727ad8159557362cf5c5359030009d4cf0aa7

cc @mmlb who found and helped debug this problem. (and tested it fixed it!)

### DIFF
--- a/nix/bogus-nixpkgs/stdenv/setup
+++ b/nix/bogus-nixpkgs/stdenv/setup
@@ -80,6 +80,11 @@ addToSearchPathWithCustomDelimiter() {
     fi
 }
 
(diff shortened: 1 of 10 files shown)


mmlb:
tested locally and confirmed to work equivalent to plain `nix-shell`

PR #112, opened 2019-07-01
Carnix

grahamc:


default.nix
```diff
@@ -2,41 +2,33 @@
...
-  '';
-
-  # Darwin fails to build doctests with:
-  # dyld: Symbol not found __ZTIN4llvm2cl18GenericOptionValueE
-  # see: https://github.com/NixOS/nixpkgs/pull/49839
-  doCheck = !pkgs.stdenv.isDarwin;
+((pkgs.callPackage ./Cargo.nix {
+  cratesIO = pkgs.callPackage ./crates-io.nix {};
+}).lorri {
+}).override {
```
tilpner: Could be single line to make empty set more obvious

Profpatsch:
That’s a *lot* of lines for a simple lockfile. Is there a deeper reason why we have to check in three different generated files to get carnix support?

grahamc:
The Cargo.nix must live in the root of the project, as the generated file always uses `./.` as the source, and I'm not interested in spending a bunch of time fixing that to avoid having the file in the root.

Profpatsch:
Great! If we add Cachix now, we can get the build times down considerably.

Should I squash-merge?

grahamc:
Sure.

PR #115, opened 2019-07-18
Reify nix expressions to cas

Profpatsch:
Best reviewed commit by commit.

This was initiated by the impossibly long nix command lines that are spawned by putting the expressions in argv.

I noticed that this change finally required the refactors I’ve been wanting to do for over a month now, so those are in here.


src/cas.rs
```diff
@@ -0,0 +1,111 @@
...
+            Err(e) => match e.kind() {
+                // since we use create_new, it will tell us if the file
+                // exists; in that case it was already written (same hash),
+                // so we don’t have to write it anew.
+                std::io::ErrorKind::AlreadyExists => Ok(()),
+                _ => Err(e),
+            },
+            Ok(mut file_handle) => {
+                file_handle.write_all(content.as_bytes())?;
+                file_handle.sync_all()
```
grahamc: If this gets interrupted, it is possible for the file to exist in the store with partial invalid content, and then never be rewritten again.

What if we did something like write to a temporary file and renamed the file in to place?

By having the temporary file in the same directory we are able to guarantee it is on the same filesystem, and thus guarantee that the `mv` is atomic.

grahamc: Another problem this write-and-move would solve is ensuring the file is only visible once it is fully written.

src/cas.rs
```diff
@@ -0,0 +1,110 @@
...
+    ///
+    /// This operation hashes the file contents, its cost is at least
+    /// the cost of hashing the `content` string.
+    pub fn file_from_string(&self, content: &str) -> std::io::Result<PathBuf> {
+        // md5 should be okay, since this is not security-critical
+        let hash = format!("{:x}", md5::compute(content.as_bytes()));
+
+        let file_name = self.store_dir.join(hash);
+
+        match std::fs::OpenOptions::new()
```
grahamc: We probably need to create a lock on the file, too.

src/cas.rs
grahamc: https://github.com/grahamc/r13y.com/blob/master/src/contentaddressedstorage.rs#L28

src/nix.rs
```diff
@@ -406,17 +434,34 @@ mod tests {
...
+            "-A",
+            "hello",
+            "--argstr",
+            "foo",
+            "bar",
             "--expr",
             "my-cool-expression",
+        ]
+        .into_iter()
+        .map(OsStr::new)
```
grahamc: These are already OsStr's though, why do we need to map them?

src/nix.rs
```diff
@@ -406,17 +434,34 @@ mod tests {
...
+        .collect();
+        assert_eq!(exp, nix.command_arguments());
+
+        let mut nix2 = CallOpts::file(PathBuf::from("/my-cool-file.nix"));
```
grahamc: This should likely be a second test, specific to file arguments

grahamc:
With the CAS store, this starts begging questions like how do we GC this store? For several of the use cases, it might be nicer to figure out how to ship files with the binary instead of inlining them to the binary. What do you think?

src/cas.rs
Profpatsch: Resolved by moving to `atomicwrites`.

src/nix.rs
Profpatsch: They’re `str`s unfortunately, rust does not change the type of a string literal depending on context.

Profpatsch:
> With the CAS store, this starts begging questions like how do we GC this store? 

Not necessary, since we only use it to reify static strings from the binary to the file system, into the user cache dir. Should not be more than a few kilobytes.

> For several of the use cases, it might be nicer to figure out how to ship files with the binary instead of inlining them to the binary. What do you think?

That’s exactly what I’m trying to prevent here, because a single binary is a lot easier to manage than a binary with a bunch of data files everywhere.

src/nix.rs
grahamc: I don't follow. The type hasn't changed, it has always been `let exp: Vec<&OsStr> = [ `, so what changed when adding a handful more arguments to require this new code?

Profpatsch:
I evaluated the `atomicwrites` module pretty in-depth before using it here, to make sure it works as intended. In the process, I submitted a documentation PR upstream, but it being merged is not a precondition for this PR.

src/nix.rs
Profpatsch: `vec!["a", "b", "c"]` is a `Vec<str>`, only with `.map(OsStr::new)` does it become a `Vec<&OsStr>` (the type is for the final result of the full iterator chain).

PR #118, opened 2019-07-26
pathreduction.rs: minor style changes

tilpner:
I feel it's a tiny bit prettier, but that's subjective, and I'm not even sure it still works (tests pass, but I didn't do any extensive testing). Feel free to close.

### DIFF
--- a/src/pathreduction.rs
+++ b/src/pathreduction.rs
@@ -29,7 +29,7 @@ impl PathReduction {
 pub fn reduce_paths(paths: &[PathBuf]) -> HashSet<PathBuf> {
     let mut reduced = paths
         .iter()
-        .map::<_, _>(|path| {
+        .map(|path| {
             let reducers = &[reduce_channel_path, reduce_nix_store_path];
 
             for reducer in reducers {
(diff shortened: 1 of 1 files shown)


src/pathreduction.rs
```diff
@@ -48,22 +48,19 @@ pub fn reduce_paths(paths: &[PathBuf]) -> HashSet<PathBuf> {
...
     // possible, in the next fold.
     reduced.sort();
+    reduced.dedup();
     reduced
         .into_iter()
         .fold::<HashSet<PathBuf>, _>(HashSet::new(), |mut set, new_path| {
-            if set.iter().any(|path| new_path.starts_with(path)) {
-                set
-            } else {
+            if !set.iter().any(|path| new_path.starts_with(path)) {
```
Profpatsch:
```suggestion
            if !set.iter().any(new_path.starts_with) {
```

tilpner: I haven't tested your suggestion, but I believe it wouldn't compile. You can eliminate the closure when there's no self, but there is here. Rust would have to implicitly create a closure that passes new_path along (but it doesn't).

Profpatsch: You are right, can’t call methods that way.

PR #122, opened 2019-08-03
Cleanup unused

justinwoo:
removes unused build command, so this is less confusing for new users.


Profpatsch:
Great, we shouldn’t advertise commands in the usage that do nothing in the end :)

PR #123, opened 2019-08-03
Install command

justinwoo:
Adds a simple install command, so that users don't need to run `lorri daemon` or `lorri watch` just to have the shell root built. This makes lorri conform to most people's expectations of how tools like this work, where you have an `init` phase that initializes config, and an `install` phase that will do the setup based on the current configuration.

This really only does the same thing as `watch` but using the `once` mode of the build loop.

Based off #122, preventing merge conflicts. Can be rebased when needed.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -54,6 +54,10 @@ pub enum Command {
     /// Bootstrap files for a new setup
     #[structopt(name = "init")]
     Init,
+
+    /// Install the current environment.
+    #[structopt(name = "install", alias = "i")]
+    Install,
 }
 
 /// Send a message with a lorri project.
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,9 @@ use lorri::locate_file;
 use lorri::NixFile;
 
 use lorri::cli::{Arguments, Command};
-use lorri::ops::{daemon, direnv, info, init, ping, shell, upgrade, watch, ExitError, OpResult};
+use lorri::ops::{
+    daemon, direnv, info, init, install, ping, shell, upgrade, watch, ExitError, OpResult,
+};
 use lorri::project::Project;
 use std::env;
 use structopt::StructOpt;
@@ -86,6 +88,10 @@ fn run_command(opts: Arguments) -> OpResult {
         Command::Ping_(p) => ping::main(p.nix_file),
 
         Command::Init => init::main(TRIVIAL_SHELL_SRC, DEFAULT_ENVRC),
+
+        Command::Install => {
+            get_shell_nix().and_then(|sn| install::main(create_project(&paths, sn)?))
+        }
     }
 }
 
--- a/src/ops/direnv/mod.rs
+++ b/src/ops/direnv/mod.rs
@@ -32,7 +32,7 @@ pub fn main(project: Project) -> OpResult {
 
     if !shell_root.exists() {
         return Err(ExitError::errmsg(
-            "Please start `lorri daemon` or run `lorri watch` before using direnv integration.",
+            "Please start `lorri daemon`, run `lorri watch`, or simply `lorri install` before using direnv integration.",
         ));
     }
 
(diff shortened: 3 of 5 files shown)


justinwoo:
Rebased this now, so it's just one small commit.

Profpatsch:
Not to bring up unnecessary bikeshedding, but what about `lorri build`?

justinwoo:
Sounds a little bit evil based off of the description of the "build" command in previous versions. I still think `install` avoids this collision and works according to how people expect `install` to install dependencies

Profpatsch:
After some deliberation, we decided we’d ideally call it `lorri run` (see commit message), but `lorri watch --once` would be the best solution for now, see commit message.

justinwoo:
You could call it "run", but having it be under a flag for watch is probably the worst one for usability. There's no way a new user will find this readily, and there's also no other tools I use where I say "watch my files to build but do it just once".

Profpatsch:
True, @grahamc should we just add the new command?

justinwoo:
I'm going to change this back in my branch because 1) I already use this 2) I have some other people also using this 3) there should be a separate command for this action. It really makes no sense to make the UI of this tool so confusing as to introduce a bunch of its own terminology.

grahamc:
Overall the code in this PR looks fine and I have no objections to the code itself. However this is not  the direction that we want to go with lorri right now. lorri is intended to be used with the watcher going continuously and if this causes any problems then we should try to tackle the actual problems and not try to come up with workarounds.

Having an option to `watch` to run just once makes sense as part of a debugging interface but making it its own top-level command implicitly makes it part of the intended workflow.

Could you maybe elaborate a bit again on your usecase scenario that made you want this command? Is adding `lorri watch --once` going to solve your problem?

justinwoo:
I don't change my shell.nix constantly, so I don't want to run a watcher or daemon to try to constantly listen for changes, ever. When introducing me to lorri, Profpatsch had me run `lorri watch` once and ctrl-C break it.

What I want to have is what lorri does currently, but with the `install` command. I don't ever want to run the daemon. When I check out or make a new project, I expect to have some static shell.nix that I will not edit for a while. When I DO start editing this shell.nix, OR it is that whatever project I'm working frequently has changes to shell.nix or related sources, I might choose to then use lorri watch or lorri daemon to actually listen to changes as I make them.

I don't see what "watch once" does other than just obfuscate using this command. I understood that lorri is supposed to be about installing dependencies for my development environment, which is why I explicitly want to "install" these once given that I am not changing them at the given moment I run this install command.

Profpatsch:
> What I want to have is what lorri does currently, but with the `install` command. I don't ever want to run the daemon.

The daemon is designed to be near-zero overhead (once we get a good grip on how many files are watched), but I can understand that people don’t want to run anything permanently that they don’t regularly use/need.

Graham is the main author of this project, so he has the final word on what he wants to be the focus and what not. I assume it’s fine with you @justinwoo if such an operation exists at all, regardless of how discoverable it is.

justinwoo:
The overhead is not an issue with me as much as knowing when and why something will change.

I am not actually fine with such a command existing "regardless of how discoverable", which is why I'm talking about this so much in here. I would really like to avoid being asked for support about something we can make readily usable now rather than later.

symbiont-joseph-kachmar:
Heyo, not an active user yet (but I’ve been looking at `lorri` as I start to explore using Nix for development a bit more).

Just wanted to say that, from a conceptual perspective, `watch once` really doesn’t make very much sense to me.

It feels as if it embraces the notion that tying some incidental behavior (installing the dependencies in a given `shell.nix`) to an almost unrelated invocation (watching the project for changes).

It makes sense that `watch` will install dependencies if nothing is installed, since the command would be useless without it, but its raison d'être _feels_ like it should be “Monitor the project and act on a detected change.”

Does that make sense?

Profpatsch:
We decided to add `lorri watch --once` for now. See https://github.com/target/lorri/pull/161

PR #117, opened 2019-07-25
builder: split nix-instantiate & nix-build

Profpatsch:
A bit of refactoring, to finally split the instantiation and build (see last commit).

Please review commit-by-commit.

- Add wrappers around `PathBuf` to distinguish drv files and final build results
- Use struct `OutputPaths` to keep results of `logged-evalatuion.nix` & move errors from use time to parse time
- remove `lorri shell`
- Fix an issue with nix output parsing
- and finally split instantiate and build

(Part of the fix for #58)

Closes #143 
Closes #50

PR #131, opened 2019-08-12
nix/CallOpts: return gc root handle

Profpatsch:
This includes the commit that removes `lorri shell`, to reduce the necessary downstream code that has to change.

Unblocks https://github.com/target/lorri/pull/117


src/nix.rs
```diff
@@ -194,22 +198,27 @@ impl CallOpts {
     /// # env::set_var("NIX_PATH", "nixpkgs=./nix/bogus-nixpkgs/");
     ///
     /// let tempdir = tempfile::tempdir().unwrap();
```
grahamc: Nice! Let's clean up the `tempdir` / `tempfile` references throughout this file.

src/nix.rs
```diff
@@ -194,22 +198,27 @@ impl CallOpts {
...
+    /// let location = location.into_os_string().into_string().unwrap();
     /// println!("{:?}", location);
     /// assert!(location.contains("/nix/store"));
     /// assert!(location.contains("hello-"));
+    /// drop(gc_root);
     /// ```
     ///
-    /// Note, path() return an error if there are multiple paths
+    /// `path` returns a handle to a temporary GC root directory
+    /// (`gc_root` in the example above). Until that is dropped,
```
grahamc: It doesn't return the directory, it returns an opaque value which keeps the paths alive.

Profpatsch:
Done, see force push.

PR #133, opened 2019-08-13
nix, builder: Fix nix output parsing

Profpatsch:
Split out from https://github.com/target/lorri/pull/117

Before, the nix output was parsed into `str`/`String` before parsing
it into lines (which is the behavior of the various `.lines()`
methods).

We can, however, not assume that nix output is valid UTF-8. Take for
example the following derivation:

```
with import <nixpkgs> {};
runCommand "foo" {} ''
  dd if=/dev/urandom of=./foo bs=1M count=1
  cat ./foo
  touch $out
''
```

which prints 1 megabyte of random output to stdout (and thus to nix’s
`stderr`). This random data is not valid UTF-8, which means lorri is
going to choke on the derivation output and crash with a decoding
error.

Instead, we keep the output a `Vec<u8>`, convert it to an
`OsString` (which is a wrapper around `Vec<u8>` on POSIX systems) and
split on `\n` (which is the line separator on POSIX systems).

As a nice side-effect, we can remove the use of threads from the
`builder.rs` code (by replacing with `wait_with_output()`) and unify
the output parsing behavior).

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -11,11 +11,9 @@ use cas::ContentAddressable;
 use regex::Regex;
 use std::any::Any;
 use std::collections::HashMap;
-use std::ffi::OsStr;
-use std::io::{BufRead, BufReader};
+use std::ffi::{OsStr, OsString};
 use std::path::PathBuf;
 use std::process::{Command, Stdio};
-use std::thread;
 use NixFile;
 
 /// Builds the Nix expression in `root_nix_file`.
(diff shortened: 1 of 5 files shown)


src/builder.rs
```diff
@@ -112,12 +88,12 @@ pub fn run(root_nix_file: &NixFile, cas: &ContentAddressable) -> Result<Info, Er
...
     Source(PathBuf),
     AttrDrv(String, PathBuf),
-    Text(String),
+    Text(OsString),
 }
 
 /// Examine a line of output and extract interesting log items in to
 /// structured data.
-fn parse_evaluation_line(line: &str) -> LogDatum {
+fn parse_evaluation_line(line: &OsStr) -> LogDatum {
```
grahamc: Instead of making this take an `&OsStr` let's make it take an `AsRef<OsStr>`, since it'll make the tests much nicer.

Here I've written a playground example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ba3c83196bc09d155e2fe40b77447c73

The key change is the signature:

```
fn parse_evaluation_line<T>(line: T) -> LogDatum
where
    T: AsRef<OsStr>,
{
```

and then:

```
    let line: &OsStr = line.as_ref();
```

after that, none of the tests should need to change.

src/nix.rs
```diff
@@ -339,6 +337,24 @@ impl CallOpts {
...
+    F: Fn(&'a OsStr) -> T,
+{
+    output
+        // We can split on \n to separate lines, because nix only runs in POSIX environments.
+        // Using `lines()` means having to convert to UTF-8 first,
+        // and nix output is not guaranteed to be valid UTF-8.
+        // (think for example a derivation that outputs random data).
+        .split(|b| &b'\n' == b)
+        .map(std::ffi::OsStr::from_bytes)
+        .filter(|s| !s.is_empty())
```
grahamc: Interesting! I think it'd be good to make a module, with a `Lines<OsStr>` iterator, and a function like:

```
fn lines(&OsStr) -> Lines<OsStr>
```

though maybe accepting a `&'a [u8]` or even `OsString` is okay too. Not sure if there is one to be preferred.

Then, the `.map().collect()` can be part of a different function, or an iter chain.

_note: don't filter out empty lines. empty lines are lines too :)_

---

After this change, the call sites of `parse_nix_output` are (mostly) just `lines(output).map(...).collect()`

grahamc:
Also, let's get exactly your sample in the PR description as a test :)

src/builder.rs
```diff
@@ -53,37 +51,15 @@ pub fn run(root_nix_file: &NixFile, cas: &ContentAddressable) -> Result<Info, Er
...
-        results.into_iter().fold(
+    let output = cmd.spawn()?.wait_with_output()?;
+
+    let stderr_results =
+        ::nix::parse_nix_output(&output.stderr, |line| parse_evaluation_line(line));
+
+    let produced_drvs = ::nix::parse_nix_output(&output.stdout, PathBuf::from);
+
+    let (paths, named_drvs, log_lines): (Vec<PathBuf>, HashMap<String, PathBuf>, Vec<OsString>) =
+        stderr_results.into_iter().fold(
```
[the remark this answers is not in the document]
grahamc: _notes from a question @Profpatsch had:_ We will want to show evaluation level output, but he's right we won't want to show the `-vvvvv` output. We'll want to show messages like:

 - copying large directories
 - other trace logs
 - errors from evaluation time

This is a bit tricky to split out. Maybe part of this future feature (streaming the logs to the user) can be implementing a more structured log to Nix, so we can filter for information we like.

---

We're going to go back to using the threads and process the log lines in a stream.

src/nix.rs
Profpatsch: I’ll leave that for a future improvement, when we really need streaming.

Profpatsch:
Pushed a WIP fixup commit, which adds a test (TODO missing comments) and addresses the comments.

src/nix.rs
Profpatsch: Was provided by the improvements @grahamc made.

Profpatsch:
This is ready for merge, right @grahamc?

PR #134, opened 2019-08-13
Builder: restrict named drvs to struct

Profpatsch:
First part of https://github.com/target/lorri/pull/117 for easier review.


src/lib.rs
```diff
@@ -81,3 +81,26 @@ impl From<PathBuf> for NixFile {
...
+    fn from(s: &std::ffi::OsStr) -> StorePath {
+        StorePath(PathBuf::from(s.to_owned()))
+    }
+}
+
+impl From<std::ffi::OsString> for StorePath {
+    fn from(s: std::ffi::OsString) -> StorePath {
+        StorePath(PathBuf::from(s))
+    }
+}
```
grahamc: Let's move StorePath in to `nix.rs`.

grahamc:
I'm not sure about using `OutputPaths` here as the name is very generic, but actually what is inside is very specific: the shell store path and derivation.

I'm also not comfortable with the type parameter for output paths, it feels like there is another abstraction trying to get out.

Thirdly, the `.map..` to add `attr-' feels super weird -- I think it is time to kill that string munging.

I think that identifying and exposing the underlying abstraction we're trying for will let us be more direct and simpler in the code, while also being more confident -- yielding simpler error handling. After lunch I  will try to propose something concrete.

Issue #116, opened 2019-07-24
`lorri direnv` should take the shell.nix file as argument


sgraf812:
I'm working on GHC, where there's https://github.com/alpmestan/ghc.nix to provision a proper build environment. I have checked that repository out in ~/ghc/nix, alongside my GHC checkouts. So, normally a build command looks like `nix-shell --pure ../nix --run "build/hadrian.sh --flavour=quick test"`. Quite a mouthful! I was hoping I could fire up lorri like `eval $(lorri direnv ../nix/default.nix)`, but apparently that doesn't work.

sgraf812:
Wait... So, ghc.nix isn't probably gonna change that frequently. So I'm probably better off with taking what `lorri watch` produces and modify it so that it works for my use case and use that as my .envrc, right?

sgraf812:
OK, `ln -s ../nix/default.nix shell.nix` seems to work for now, at least it builds something in ~/.cache. But apparently although I get "direnv: loading .envrc" when entering the directory in my zsh shell, The environment is rather the same as before; None of the build dependencies are present. Weird.

Profpatsch:
Internally, I refactored everything to not depend on `shell.nix` anymore, so we can add a `--shell-file` option to `lorri watch`/`lorri direnv` at any time.

jfroche:
merged, this issue can be fixed

PR #136, opened 2019-08-13
feat: add option to define the shell.nix location

jfroche:
use --shell-file or -s option to declare the nix file to use when
running command info, watch, direnv

refs #116

:warning:  first time I contribute to a rust project.

thank you for this great project, it saves me a lot of time
thank you for marking this as good first issue

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -25,15 +25,15 @@ pub enum Command {
     /// Emit shell script intended to be evaluated as part of
     /// direnv's .envrc, via: `eval "$(lorri direnv)"`
     #[structopt(name = "direnv")]
-    Direnv,
+    Direnv(NixShellFile),
 
     /// Show information about the current Lorri project
     #[structopt(name = "info", alias = "information")]
-    Info,
+    Info(NixShellFile),
 
     /// Build `shell.nix` whenever an input file changes
     #[structopt(name = "watch")]
-    Watch,
+    Watch(NixShellFile),
 
     /// Start the multi-project daemon. Replaces `lorri watch`
     #[structopt(name = "daemon")]
@@ -62,6 +62,14 @@ pub struct Ping_ {
     /// The .nix file to watch and build on changes.
     #[structopt(parse(from_os_str))]
     pub nix_file: NixFile,
+}
+
+///  Add parameter to define the shell file in the current directory to use
+#[derive(StructOpt, Debug)]
+pub struct NixShellFile {
+    /// The .nix file in the current directory to use
+    #[structopt(long = "shell-file", parse(from_os_str), default_value = "shell.nix")]
+    pub nix_file: PathBuf,
 }
 
 /// A stub struct to represent how what we want to upgrade to.
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,7 @@ use lorri::NixFile;
 use lorri::cli::{Arguments, Command};
 use lorri::ops::{daemon, direnv, info, init, ping, upgrade, watch, ExitError, OpResult};
 use lorri::project::Project;
-use std::env;
+use std::path::PathBuf;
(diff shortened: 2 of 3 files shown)


src/cli.rs
```diff
@@ -64,6 +64,19 @@ pub struct Ping_ {
     pub nix_file: NixFile,
 }
 
+///  Add parameter to define the shell file to use
+#[derive(StructOpt, Debug)]
+pub struct NixShellFile {
+    /// The .nix file to use
+    #[structopt(
+        short = "s",
```
grahamc:
```suggestion
```

src/cli.rs
```diff
@@ -64,6 +64,19 @@ pub struct Ping_ {
     pub nix_file: NixFile,
 }
 
+///  Add parameter to define the shell file to use
+#[derive(StructOpt, Debug)]
+pub struct NixShellFile {
+    /// The .nix file to use
```
grahamc: Because we use `locate_file` when loading the file, the documentation should say that it must be a single file's name, and not be `../foo.nix` or `./bar/baz.nix` etc.

grahamc:
At first read, this looks really good. I made a couple small suggestion, does that seem okay to you? Since nobody will type it (except for once per project) I think it is nicer to spell it out fully.

src/cli.rs
kyren: Does that mean we won't be able to disambiguate between, say, a shell.nix file in a project subdirectory and a shell.nix file in a git submodule of that project?

src/main.rs
```diff
@@ -41,18 +42,23 @@ fn main() {
...
-    Ok(NixFile::from(locate_file::in_cwd("shell.nix").map_err(
+    let shellfile_str = match shellfile.to_str() {
+        Some(i) => i,
+        _ => "",
+    };
+    Ok(NixFile::from(locate_file::in_cwd(shellfile_str).map_err(
         |_| {
             ExitError::errmsg(format!(
-                "There is no `shell.nix` in the current directory{}\n\
+                "There is no `{}` in the current directory{}\n\
```
Profpatsch: It also works with an absolute path like `/home/user/shell.nix`.

Then the error message would say `"There is no `/home/user/shell.nix` in the current directory"`.

Instead, when the user provides a `shellfile`, it should always print a different message, like `"{} does not exist.`.

src/main.rs
```diff
@@ -41,18 +42,23 @@ fn main() {
...
         Err(_) => String::from(""),
         Ok(pb) => format!(" ({})", pb.display()),
     };
     // use shell.nix from cwd
-    Ok(NixFile::from(locate_file::in_cwd("shell.nix").map_err(
+    let shellfile_str = match shellfile.to_str() {
+        Some(i) => i,
+        _ => "",
+    };
+    Ok(NixFile::from(locate_file::in_cwd(shellfile_str).map_err(
```
Profpatsch: `shellfile.to_str()` might return an error if the given string is not valid UTF-8. We should change `locate_file::in_cwd()` to work on `PathBuf`s or `OSString` instead (so we don’t have to decode to UTF-8 unnecessarily).
If you don’t know how to do that, I can do it afterwards if you want.

jfroche: Should be done now. 
But we still need to use shellfile.to_str() to show the failing path (in case of error)

Profpatsch: You can use https://doc.rust-lang.org/std/path/struct.Path.html#method.display, which is defined for `Path` and thus also for `PathBuf` (`Path` is the borrowed version of `PathBuf`).

jfroche: Learnt something, thank you. Should be fixed now.

jfroche:
hello, is there something missing ?

Profpatsch:
Was on vacation, LGTM

PR #137, opened 2019-08-15
Graham/fix nix output parsing

grahamc:
The code in #133 is pretty good, check out these proposed patches for some feedback / suggestions

This borrows upstream Rust code to implement a Lines iterator which produces OsStrings, and replaces the parse_nix_output code to use it. From the commit where I drop that function:

> nix: drop parse_nix_output
>
> Prefer the iterator format, especially when we're able to
> use familiar forms of passing functions around.
>
> In this case the parse_nix_output function probably shouldn't
> skip newlines.
>
> Additionally, reducing the powerful model of iterators down
> to only supporting (and actually requiring) a single function
> isn't doing us much favor.

Finally, in the tests for the builder I add a proof-positive test to assert that the bytes come through and that the shell derivation is produced.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -8,6 +8,7 @@
 //! `stderr`, like which source files are used by the evaluator.
 
 use cas::ContentAddressable;
+use osstrlines;
 use regex::Regex;
 use std::any::Any;
 use std::collections::HashMap;
@@ -66,8 +67,9 @@ pub fn run(root_nix_file: &NixFile, cas: &ContentAddressable) -> Result<Info, Er
 
     let stderr_results: thread::JoinHandle<std::io::Result<Vec<LogDatum>>> =
         thread::spawn(move || {
-            let reader = BufReader::new(stderr);
-            ::nix::parse_nix_output(reader, |line| parse_evaluation_line(&line))
+            osstrlines::Lines::from(BufReader::new(stderr))
+                .map(|line| line.map(parse_evaluation_line))
+                .collect::<Result<Vec<LogDatum>, _>>()
         });
 
     let produced_drvs: thread::JoinHandle<std::io::Result<Vec<PathBuf>>> =
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -211,6 +211,7 @@ mod tests {
     use super::*;
     use cas::ContentAddressable;
     use std::ffi::OsString;
+    use std::os::unix::ffi::OsStrExt;
     use std::path::PathBuf;
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -256,13 +263,11 @@ derivation {{
   allowSubstitutes = false;
   preferLocalBuild = true;
   system = builtins.currentSystem;
-  # this is to make nix rebuild for every test, I’m so sorry
-  random = {:?};
+  # this is to make nix rebuild for every test
+  random = builtins.currentTime;
```
Profpatsch: I’m not sure this is better than the tempdir, mainly because I’m afraid future versions of nix will stub that out to be always the same in pure mode (c.f. flakes)

src/builder.rs
```diff
@@ -290,6 +295,15 @@ in {}
 
         let info = run(&::NixFile::from(cas.file_from_string(&nix_drv)?), &cas).unwrap();
         assert!(info.exec_result.success());
+        assert!(info
+            .named_drvs
+            .get("shell")
+            .unwrap()
+            .to_string_lossy()
+            .contains("-shell.drv"));
```
Profpatsch: I’d pull that out into a different test (if we really want/need to test it, the code itself (after #117) is going to have the asserts and will fail if it’s missing).

src/builder.rs
```diff
@@ -290,6 +295,15 @@ in {}
...
+
+        let expect: OsString = OsStr::from_bytes(b"\"\xAB\xBC\xCD\xDE\xDE\xEF\"").to_owned();
+        assert!(info.log_lines.contains(&expect));
```
Profpatsch: Good idea!

src/nix.rs
```diff
@@ -296,7 +295,9 @@ impl CallOpts {
 
         if output.status.success() {
             let stdout: &[u8] = &output.stdout;
-            let paths: Vec<PathBuf> = ::nix::parse_nix_output(stdout, PathBuf::from)?;
+            let paths: Vec<PathBuf> = osstrlines::Lines::from(stdout)
+                .map(|line| line.map(PathBuf::from))
+                .collect::<Result<Vec<PathBuf>, _>>()?;
```
Profpatsch: This is a lot better, putting the function was very much fighting the system. I wasn’t sure how much work an iterator would be, but it seems to be totally reasonable.

src/osstrlines.rs
```diff
@@ -0,0 +1,73 @@
...
+    type Item = Result<OsString>;
+
+    fn next(&mut self) -> Option<Result<OsString>> {
+        let mut buf = vec![];
+        match self.buf.read_until(b'\n', &mut buf) {
+            Ok(0) => None,
+            Ok(_n) => {
+                if buf.ends_with(&[b'\n']) {
+                    buf.pop();
+                    if buf.ends_with(&[b'\r']) {
```
Profpatsch: Do we need to think about '\n\r`? Nix doesn’t run on Windows (outside of WSL), and MacOS has switched to `\n` when they went to BSD.

gilligan: The implementation is close to verbatim from the `lines` function of `bufread` so it was decided to leave it as it is.

PR #146, opened 2019-09-05
Builder default.nix logic

Profpatsch:
Finally solves the issue of watching way too many files. See commit messages for details.

<s>
Merge #117 first!
Edit: the utf8-test will break, because it’s still broken on #117.
</s>
Is now based on `master`.


Profpatsch:
@grahamc This is based on master now and can be reviewed independently.

PR #147, opened 2019-09-05
Fixed a Typo in the Travis cfg

teofr:
The `.travis.yml.nix` was diffing `.travis.yml` under the `carnix-idempotent` travis step, this is probably a typo and should be `Cargo.nix`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@
       lorri_travis_fold travis-yml-idempotent \
         git diff -q ./.travis.yml
       lorri_travis_fold carnix-idempotent \
-        git diff -q ./.travis.yml
+        git diff -q ./Cargo.nix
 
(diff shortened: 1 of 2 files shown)


Profpatsch:
Actually I’m not exactly sure what this checks for in the first place.

The travis idempotence check runs `nix-instantiate` in the lint and checks whether that does anything, but we don’t run `carnix generate-nix` in the lint (which takes a while), so checking for idempotence can never fail if I see this correctly.

teofr:
Isn't `carnix generate-nix` being run from here https://github.com/target/lorri/blob/b5c089aaf27fe3f31ade7f4b21264c6aa2adfcd5/shell.nix#L92-L93

Profpatsch:
You are, of course, absolutely right.

Issue #120, opened 2019-07-30
When BuildLoop hits an unrecoverable failure, lorri daemon should exit


chreekat:
Current behavior:

An unrecoverable failure (e.g. #66) kills the thread running the loop, but the lorri daemon keeps on truckin'. But it doesn't do anything. This means that a daemonized daemon must be babysat and restarted manually when hitting some unresolved "file not found" error over and over again.

Expected behavior:

If the looper thread perishes, the daemon should exit (or possibly restart the thread, but I vote for just dying and letting the system handle it).

chreekat:
I should mention, when I said "e.g. #66", I mean my daemon is erroring out at the exact same call to unwrap(), as pasted below. To be clear, I don't give a hoot about the particular `Err` value, just the way the daemon handles it.


```
Jul 30 14:03:35 bryan-laptop lorri[7914]: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Unrecoverable(Notify(Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })))', src/libcore/result.rs:997:5
Jul 30 14:03:35 bryan-laptop lorri[7914]: stack backtrace:
Jul 30 14:03:35 bryan-laptop lorri[7914]:    0:     0x560ae72926df - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::hfad0220813d5e5c3
Jul 30 14:03:35 bryan-laptop lorri[7914]:    1:     0x560ae72a2d67 - std::sys_common::backtrace::print::h16017ddc8b354e39
Jul 30 14:03:35 bryan-laptop lorri[7914]:    2:     0x560ae72960a3 - std::panicking::default_hook::{{closure}}::h8b06d3b8a7138cc5
Jul 30 14:03:35 bryan-laptop lorri[7914]:    3:     0x560ae7295e23 - std::panicking::default_hook::h8cac4830452b35df
Jul 30 14:03:35 bryan-laptop lorri[7914]:    4:     0x560ae729679f - std::panicking::rust_panic_with_hook::hd32c26e1600a179b
Jul 30 14:03:35 bryan-laptop lorri[7914]:    5:     0x560ae7296331 - std::panicking::continue_panic_fmt::h24b78f7279f44cef
Jul 30 14:03:35 bryan-laptop lorri[7914]:    6:     0x560ae7296215 - rust_begin_unwind
[...]

PR #148, opened 2019-09-11
daemon, build_loop, watch: crash on unrecoverable error

Profpatsch:
Previously, we panic!()’ed in the build_loop, inside of a
forever-Thread, meaning the thread would die but the daemon continues
running.

This is solved by forwarding the message throught the existing channel
and handling it in the main thread.

Tested manually by stubbing out `run` in `builder` and verifying that the daemon crashes.

Closes: https://github.com/target/lorri/issues/120
cc @chreekat


grahamc:
I implemented this a bit differently in #156, by using a Drop implementation. The main advantage is even unplanned crashes will still take the daemon down completely, so in the case of major problems a panic still properly kills the daemon.

PR #156, opened 2019-09-24
daemon: panic if any threads panic

grahamc:
Alternate implementation to #148 

This implementation handles any panic in the spawned threads, and doesn't require explicitly returning a specific unrecoverable error type.

Also, it gets rid of a few todo's in the daemon.

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -1,0 +1,106 @@
+//! A thread pool for panicking immediately after any monitored threads
+//! panic.
+//!
+//! The key implementation detail is each thread spawned gets a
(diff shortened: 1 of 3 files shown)


src/thread.rs
```diff
@@ -0,0 +1,106 @@
...
+
+            let handle = self
+                .threads
+                .remove(&thread_id)
+                .expect("Failed to find thread ID in handle table");
+
+            handle
+                .join()
+                .expect("Failed to join thread, despite catch_unwind!");
+        }
```
Profpatsch: Where is the actual forwarding of the panic of a panicing thread happening? I don’t see it.

PR #157, opened 2019-09-26
Bug 151/start lorri daemon

### LINKED ISSUE #113: `direnv allow` asks to start watcher/daemon when project has not yet built
When enabling an .envrc that uses lorri for some shell.nix which has not yet actually been executed, lorri will output:

```
Please start `lorri daemon` or run `lorri watch` before using direnv integration.
```

even when the lorri daemon is running. This seems to happen only when a build has not yet completed for the relevant directory, which makes sense. lorri is probably only checking for some cached nix shell and because it does not exist it decides to output this message. The behavior is still likely to be confusing for the end user though.

I'm not sure what the best thing to do in this scenario is (Is there an easy way to talk to the daemon and ask if it has a build current taking place for the current directory?).

### LINKED ISSUE #151: "Please start `lorri daemon`", but it's already running
On my system, `lorri direnv` is complaining that the daemon isn't running, even though it is (started as a systemd service). It also seems to recognize this, because `lorri daemon` fails with 

> Another daemon is already listening on the socket at /run/user/1000/lorri/daemon.socket. We are currently only allowing one daemon to be running at the same time.

This seems weird, maybe `lorri direnv` should output more info so the error can be diagnosed easier?

grahamc:
Fix #151
Fix #113 

1. Print a more correct warning to the user if the environment is not cached, or if the lorri daemon is not updated.
2. Simplify handling when the daemon isn't running by watching the socket, so we evaluate when the daemon starts.

### DIFF
--- a/src/ops/direnv/mod.rs
+++ b/src/ops/direnv/mod.rs
@@ -17,25 +17,45 @@ pub fn main(project: Project) -> OpResult {
 
     let socket_path = ::ops::get_paths()?.daemon_socket_file().to_owned();
 
-    // TODO: don’t start build/evaluation automatically, let the user decide
-    if let Ok(client) =
-        client::ping(DEFAULT_READ_TIMEOUT).connect(&::socket::path::SocketPath::from(&socket_path))
-    {
+    let root_paths = Roots::from_project(&project).paths();
+    let paths_are_cached: bool = root_paths.all_exist();
+
+    let ping_sent: bool = if let Ok(client) = client::ping(DEFAULT_READ_TIMEOUT).connect(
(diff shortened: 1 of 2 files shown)


src/ops/direnv/mod.rs
```diff
@@ -46,11 +48,16 @@ pub fn main(project: Project) -> OpResult {
...
+watch_file "{}"
 watch_file "$EVALUATION_ROOT"
 
 {}
 "#,
         root_paths.shell_gc_root,
+        socket_path
+            .into_os_string()
+            .into_string()
+            .expect("Socket path is not UTF-8 clean!"),
```
Profpatsch: That’s clever!

PR #161, opened 2019-10-02
add `lorri watch --once`

Profpatsch:
Adds a `--once` argument to lorri watch, to only run one evaluation.

`lorri install` was proposed instead, but we don’t want it to sound
like the command changes the system of the user (think `make
install`).

Ultimately, `lorri watch` is going to be replaced by `lorri daemon`
completely, then we should introduce a `lorri run` command which does
the same as `lorri watch --once`.

Co-Authored-By: Graham Christensen <graham@grahamc.com>
Co-Authored-By: Tobias Pflug <tobias.pflug@tweag.io>

### DIFF
--- a/src/ops/watch.rs
+++ b/src/ops/watch.rs
@@ -1,28 +1,53 @@
 //! Run a BuildLoop for `shell.nix`, watching for input file changes.
 //! Can be used together with `direnv`.
-use crate::build_loop::BuildLoop;
-use crate::ops::{ok, OpResult};
+use crate::build_loop::{BuildError, BuildLoop};
+use crate::cli::WatchOptions;
+use crate::ops::{ok, ExitError, OpResult};
 use crate::project::Project;
+use std::fmt::Debug;
+use std::io::Write;
 use std::sync::mpsc::channel;
 use std::thread;
 
+// TODO: this should probably be something more structured
+fn print_build_message<A>(msg: A)
+where
+    A: Debug,
+{
+    println!("{:#?}", msg);
+}
+
 /// See the documentation for lorri::cli::Command::Shell for more
 /// details.
-pub fn main(project: Project) -> OpResult {
+pub fn main(project: Project, opts: WatchOptions) -> OpResult {
     let (tx, rx) = channel();
 
-    let build_thread = {
-        thread::spawn(move || {
-            let mut build_loop = BuildLoop::new(&project);
-            build_loop.forever(tx);
-        })
-    };
+    if opts.once {
+        let mut build_loop = BuildLoop::new(&project);
+        match build_loop.once() {
+            Ok(msg) => {
+                print_build_message(msg);
+                ok()
+            }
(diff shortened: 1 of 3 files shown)


src/ops/mod.rs
```diff
@@ -55,7 +55,7 @@ impl ExitError {
     /// Helpers to create exit results
     ///
     /// Note: err panics if exitcode is zero.
-    fn err<T>(exitcode: i32, message: T) -> ExitError
+    pub fn err<T>(exitcode: i32, message: T) -> ExitError
```
grahamc: Let's make a function which creates an error of exitcode 100, with a variable message, named to represent what that error code represents.

src/ops/watch.rs
```diff
@@ -1,28 +1,42 @@
...
+                build_loop.forever(tx);
+            })
+        };
 
-    for msg in rx {
-        println!("{:#?}", msg);
-    }
+        for msg in rx {
+            println!("{:#?}", msg);
+            let _ = std::io::stdout().flush();
```
grahamc: Tell me about this?

src/ops/watch.rs
```diff
@@ -1,28 +1,42 @@
...
     let (tx, rx) = channel();
 
-    let build_thread = {
-        thread::spawn(move || {
-            let mut build_loop = BuildLoop::new(&project);
-            build_loop.forever(tx);
-        })
-    };
+    if opts.once {
+        let mut build_loop = BuildLoop::new(&project);
```
grahamc: Let's move the `once` codepath in to a function (inside this same file) called `main_run_once`, and move the `forever` codepath in to a function (again, same file) called `main_run_forever`, and switch between them here in `main.

src/ops/mod.rs
Profpatsch: Not actually using this in this commit, leftover from splitting out of the other PR probably. Will remember it when the merge conflicts hit.

src/ops/watch.rs
Profpatsch: Split into different commit.

Profpatsch:
@grahamc Addressed suggestions, please see force push and new commits.

src/ops/watch.rs
```diff
@@ -19,10 +43,19 @@ pub fn main(project: Project) -> OpResult {
...
+        print_build_message(msg);
+        let _ = std::io::stdout().flush();
     }
 
     build_thread.join().unwrap();
 
     ok()
 }
+
+// TODO: this should probably be something more structured
```
grahamc: Let's drop the TODO. We'll know it needs to be TO-DONE when we get to the feature:

```suggestion
```

src/ops/watch.rs
```diff
@@ -19,10 +43,19 @@ pub fn main(project: Project) -> OpResult {
     };
 
     for msg in rx {
-        println!("{:#?}", msg);
+        print_build_message(msg);
+        let _ = std::io::stdout().flush();
```
grahamc:
```suggestion
        let _ = std::io::stdout().flush();
```

Should this be part of `print_build_message`?

Issue #140, opened 2019-08-17
IN_NIX_SHELL is set to 1 instead of "pure" or "impure"


bbigras:
I'm trying to add nix-shell support to a prompt tool and it seems lorri sets IN_NIX_SHELL to 1 instead of "pure" or "impure".

Someone on irc said that it should be set to "pure".

Profpatsch:
```
> nix-shell -p --command 'printenv IN_NIX_SHELL'
impure
> nix-shell -p --pure --command 'printenv IN_NIX_SHELL'
pure
```

This must have changed recently, because I can only ever remember it being `1`.

PR #164, opened 2019-10-03
Set IN_NIX_SHELL to impure (#140)

johnchildren:
Sets IN_NIX_SHELL environment variable to `impure` instead of 1.

I think this should resolve https://github.com/target/lorri/issues/140 as I couldn't see any other references to IN_NIX_SHELL, but I should probably add a test as well? I checked it manually by checking the value of the environment variable with direnv/lorri in the development shell for this project.


grahamc:
Thank you!

PR #166, opened 2019-10-04
builder: add test for default.nix logic

Profpatsch:
The part of https://github.com/target/lorri/pull/146 that depends on #117 to be merged, that is: the test.

We shouldn’t emit any paths to the watcher that are not strictly
necessary; the default.nix logic ensures that we only watch the
special `default.nix` file if a directory is imported.
This test checks that that is actually the case.

In the future it should be extended for other cases of watching too
much, once we discover them.


Profpatsch:
Rebased on master

PR #168, opened 2019-10-11
Lorri upgrade refactor and branch subcommand

Profpatsch:
Adds `lorri self-upgrade branch` and makes the parsing more solid.


src/ops/upgrade.nix
```diff
@@ -0,0 +1,43 @@
...
+
+  fetchBranch = branch: fetchGit {
+      url = "https://github.com/target/lorri.git";
+      ref = branch;
+  };
+
+  fetchedSource =
+    if      type == "branch" then fetchBranch branch
+    else if type == "local" then path
+    else abort "impossibru";
```
curiousleo: Cute, but not very helpful!

Profpatsch: easy to grep for :)

Profpatsch:
> * If I'm a sophisticated user living on the bleeding edge, I'll figure out that I need to `git clone ; cargo build` easily enough.

We have users who would like to use lorri without going through the indirection of nixpkgs (and who don’t want to fetch a whole nixpkgs just to update lorri). Having lorri in nixpkgs is a nice addition, but it’s way easier for us to tell people “run `lorri self-upgrade …` to check out this patch”).

gilligan:
This hasn't been updated in a very long time. Personally i'm a bit in doubt about this and if this should be included in lorri. 

The PR description doesn't add any motivation unfortunately but I think `self-upgrade` works as intended and i'm not sure why to extend this in any way. I'd rather opt for keeping it simple. There is one way of upgrading and one source of truth and that's it.

Beyond that it's open source and people can build whatever branch or revision they want to.

Profpatsch:
The motivation here is to make it easy for users to try out a branch that devs push, so that they can verify whether it fixes their problems.

Instead of “check out the repo and run these nix commands” we just tell them “run `lorri self-upgrade branch this-should-fix-your-problem”

Profpatsch:
Since it’s a rather simple incremental change over our current self-upgrade command with a clear improvement for user interaction, I expected it to be fairly uncontroversial back when I submitted it. But maybe I should have given a better motivation indeed.

Profpatsch:
fixes #446

codygman:
I can say that I have a lot of bugs to respond to using emacs, nix, direnv, ghcide, etc. I also have a high tolerance for thinking with broken things or getting stuck am hour debugging a few simple commands that "should have worked".

At the time I ran the lorri upgrade command to test because it was so easy to do.

I think it makes it a lot more likely for other users to participate in testing that don't have as high of a tolerance for tinkering.

src/ops/upgrade.rs
```diff
@@ -0,0 +1,160 @@
...
+                .arg("--install")
+                .arg(build_result.as_path())
+                .status()
+                // TODO: check existence of commands at the beginning
+                .expect("Error: failed to execute nix-env --install");
+            // we can drop the temporary gc root
+            drop(gc_root);
+
+            if status.success() {
+                Ok(Some(String::from("\nUpgrade successful.")))
```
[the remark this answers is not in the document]
Profpatsch: Has been fixed on master.

Profpatsch:
I rebased and addressed the remaining comments.

I’m gonna merge it, whether or what part of the module we want to remove is for a different PR.

PR #170, opened 2019-10-14
CI: move checks into nix derivations

Profpatsch:
This moves our CI checks into scripts generated by nix.

The big idea here is to achieve maximum independence from the CI environment the tests are run in. Running

```
nix-shell -A ci.testsuite shell.nix
```

should always give *exactly* the same test results as on CI (minus architecture differences).
This is achieved by carefully cleaning the environment of every test before we execute it.

This would be nearly impossible to do with normal shell scripts, but nix’s ability to generate small utility scripts on-the-fly plays wonderfully with the idea behind CI checks. Moreover, using the `execline` language enables us to embed scripts into nix lists, so string concatenation problems and string escaping problems are a thing of the past. The resulting DSL is extremely powerful in its abstraction capabilities, while nearly trivial in implementation.


gilligan:
This hasn't moved forward in any way - @curiousleo @Profpatsch - what are your thoughts on this?

Are there **(a)** shortcomings with the current CI setup that need to be fixed ? and would **(b)** the changes in this PR help to address those?

if neither applies i'd probably suggest closing the PR

Profpatsch:
I think the general approach is valid, it can probably be simplified a bit and split into smaller PRs.

Profpatsch:
Rebased & merged in https://github.com/nix-community/lorri/pull/23/

Issue #49, opened 2019-04-04
Display reason for evaluation


Profpatsch:
We should make it clear to the user why every evaluation we use happens. This also helps debugging.

I envision this as our watching module exporting a `Reason` type with detailed information about why the change was triggered, which we display before the evaluation.

PR #171, opened 2019-10-15
Report cause of build

nyarly:
This addresses #49 - The watch module gets a `Reason` type, and reports `FilesChanged(path, count)`. This could certainly be iterated on for friendliness, but right now it's a step up from just `Started`.

### DIFF
--- a/src/build_loop.rs
+++ b/src/build_loop.rs
@@ -8,7 +8,7 @@ use crate::pathreduction::reduce_paths;
 use crate::project::roots;
 use crate::project::roots::Roots;
 use crate::project::Project;
-use crate::watch::Watch;
+use crate::watch::{Reason, Watch};
 use std::path::PathBuf;
 use std::sync::mpsc::{channel, Sender};
 
@@ -16,7 +16,7 @@ use std::sync::mpsc::{channel, Sender};
 #[derive(Clone, Debug)]
 pub enum Event {
     /// The build has started
-    Started,
+    Started(Reason),
     /// The build completed successfully
     Completed(BuildResults),
     /// The build command returned a failing exit status
@@ -64,12 +64,13 @@ impl<'a> BuildLoop<'a> {
     /// When new filesystem changes are detected while a build is
     /// still running, it is finished first before starting a new build.
     pub fn forever(&mut self, tx: Sender<Event>) {
+        let mut reason = Reason::ProjectAdded(self.project.nix_file.clone());
         loop {
             // TODO: Make err use Display instead of Debug.
             // Otherwise user errors (especially for IO errors)
             // are pretty hard to debug. Might need to review
             // whether we can handle some errors earlier than here.
-            tx.send(Event::Started)
+            tx.send(Event::Started(reason))
                 .expect("Failed to notify a started evaluation");
 
             match self.once() {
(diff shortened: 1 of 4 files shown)


src/watch.rs
```diff
@@ -43,22 +52,27 @@ impl Watch {
...
     /// Block until we have at least one event
-    pub fn block(&mut self) -> Result<(), ()> {
-        if self.blocking_iter().next().is_none() {
-            debug!("No event received!");
-            return Err(());
+    pub fn block(&mut self) -> Result<Reason, ()> {
+        match self.blocking_iter().next() {
+            Some(event) => {
+                let extra = self.process_ready();
+                Ok(Reason::FilesChanged(event.path.ok_or(())?, extra?))
```
Profpatsch: I noticed that `process_ready` does not actually return an `Err`, so you can change that to a plain `u16` and remove the `?` on `extra`.

src/build_loop.rs
```diff
@@ -64,27 +65,42 @@ impl<'a> BuildLoop<'a> {
...
-            // TODO: Make err use Display instead of Debug.
-            // Otherwise user errors (especially for IO errors)
-            // are pretty hard to debug. Might need to review
-            // whether we can handle some errors earlier than here.
-            tx.send(Event::Started)
-                .expect("Failed to notify a started evaluation");
+        let send = |msg| tx.send(msg).expect("Failed to send an event");
+
+        send(Event::Started(Reason::ProjectAdded(
+            self.project.nix_file.clone().0,
```
grahamc: hmm tuple access like this always makes me regret using tuples :P

src/build_loop.rs
```diff
@@ -64,27 +65,42 @@ impl<'a> BuildLoop<'a> {
...
     /// still running, it is finished first before starting a new build.
     pub fn forever(&mut self, tx: Sender<Event>) {
-        loop {
-            // TODO: Make err use Display instead of Debug.
-            // Otherwise user errors (especially for IO errors)
-            // are pretty hard to debug. Might need to review
-            // whether we can handle some errors earlier than here.
-            tx.send(Event::Started)
-                .expect("Failed to notify a started evaluation");
+        let send = |msg| tx.send(msg).expect("Failed to send an event");
```
grahamc: I think it'd helpful to print out the msg which failed to send

src/watch.rs
```diff
@@ -43,24 +77,30 @@ impl Watch {
...
 
     /// Block until we have at least one event
-    pub fn block(&mut self) -> Result<(), ()> {
-        if self.blocking_iter().next().is_none() {
-            debug!("No event received!");
-            return Err(());
+    pub fn block(&mut self) -> Result<Reason, RawEventError> {
+        match self.blocking_iter().next() {
+            Some(event) => {
+                let extra = self.process_ready();
```
grahamc: `extra` is really a count of changed events, right?

src/watch.rs
```diff
@@ -16,6 +16,40 @@ pub struct Watch {
...
+}
+
+/// Description of the project change that triggered a build.
+#[derive(Clone, Debug)]
+pub enum Reason {
+    /// When a project is presented to Lorri to track, it's built for this reason.
+    ProjectAdded(PathBuf),
+    /// When there is a filesystem change, the first changed file is recorded,
+    /// along with a count of other filesystem events.
+    FilesChanged(PathBuf, NumberOfBatchedEvents),
```
grahamc: Why only a single PathBuf, instead of a list of all path changes we saw?

nyarly:
Several good notes here, let me see if I can't turn those around quickly.

src/watch.rs
nyarly: My experience with inotify has been that events come in big and strange batches. I'd wanted to try it out for a little while before unwrapping "and 3 other files" into the list., in case it was often "and 300 other files." If that's not a worry (i.e. we're okay with 300 lines in log files if that happens), that's easy enough to change to.

Profpatsch:
Thanks, cleaned up the commits and rebased on master. Merging when green.

PR #172, opened 2019-10-15
Creating a coverage script based on kcov

nyarly:


Profpatsch:
Hm, what is this for?

nyarly:
I was trying to determine what code was covered by existing tests. `kcov` does the job rather nicely, and produces output like

Selection_014

I thought it was useful and might be worth sharing. Feel free to decline the PR, though.

Profpatsch:
Hm, I’m not sure simple coverage statistics like that are particularly informative.

I’m pretty sure though we are testing most interesting paths; sometimes writing a test takes an unreasonable amount of time, or is just not feasible (e.g. for path handling logic or the executables in `src/ops`).

nyarly:
As a new contributor, it was very useful to me to be able to see an objective view on how those decisions had been made in the past. It helps inform where I should spend time adding tests for new code.

On other projects, I've more than once found that the indifferent reporting of a coverage tool helped identify code paths that I was sure were tested but had been overlooked.

As I said, it was an interesting exercise (and the ease of setting up coverage was gratifying), but if it's not useful to the project, please do decline the PR.

PR #174, opened 2019-10-15
src/nix,builder: return more descriptive error if nix not in PATH

Profpatsch:


Profpatsch:
This has to be rebased onto the new `execute` function in `builder.nix`.

grahamc:
Looks good post rebase.

Profpatsch:
I opened https://github.com/rust-lang/rust/issues/65968 so maybe we can get rid of enums like this in the future.

Profpatsch:
@grahamc I had to introduce two error wrappers when rebasing onto the new code, PTAL.

PR #177, opened 2019-10-16
nix.rs: allow reading stderr output via Rx/Tx

grahamc:
Callers can pass a `Sender<OsString>` to CallOpts. Each line of stderr
outpuut from Nix is passed over this channel.

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -380,14 +380,9 @@ impl CallOpts {
         cmd.stdout(Stdio::piped());
 
         // 0. spawn the process
-        // 1. spawn a stderr handling thread
-        // 2. spawn a stdout handling thread (?)
-        // 3. wait on the process
-        // 4. join the stderr handler
-        // 5. join the stdout handler
-
         let mut nix_proc = cmd.spawn()?;
 
+        // 1. spawn a stderr handling thread
         let stderr_handle: ChildStderr = nix_proc.stderr.take().expect("failed to take stderr");
         let stderr_tx = self.stderr_line_tx.clone();
         let stderr_thread = thread::spawn(move || {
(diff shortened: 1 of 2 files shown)


src/nix.rs
```diff
@@ -110,9 +115,66 @@ impl CallOpts {
...
+    ///         .unwrap()
+    ///         ;
+    ///
+    /// let messages: Vec<String> = rx.iter()
+    ///     .map(|osstr| osstr.to_string_lossy().into())
+    ///     .collect();
+    /// println!("messages: {:#?}", messages);
+    /// assert!(messages.contains(&String::from("these derivations will be built:")), "looking for 'these derivations will be built:'");
+    /// assert!(messages.iter().any(|m| m.contains("hello-unstable-1.0.0.drv")), "looking for a line like '  /nix/store/...-hello-unstable-1.0.0.drv'");
+    /// assert!(messages.iter().any(|m| m.contains("building '/nix/store")), "looking for a line like 'building \'/nix/store...'");
```
Profpatsch: This is a bit long for an example, maybe put that in a test?

src/nix.rs
```diff
@@ -313,25 +378,73 @@ impl CallOpts {
...
+    {
+        cmd.stderr(Stdio::piped());
+        cmd.stdout(Stdio::piped());
+
+        // 0. spawn the process
+        // 1. spawn a stderr handling thread
+        // 2. spawn a stdout handling thread (?)
+        // 3. wait on the process
+        // 4. join the stderr handler
+        // 5. join the stdout handler
```
Profpatsch: These could be interspersed with the code below.

grahamc:
Addressed all the feedback, thanks!

Issue #114, opened 2019-07-10
Include human-panic or similar


Profpatsch:
https://github.com/rust-cli/human-panic

`human-panic` captures panics, prints them (and more information) to a file and prints a message which asks the user to submit an issue with the contents of the panic file

It has a few dependencies I’m not sure about, but we should include the gist of it into lorri.

johnchildren:
Looks like it might be possible to replicate some of this functionality directly using the standard library's [set_hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) as listed in their README. This way it would be possible to have a more customized message without the extra dependencies (potentially).

I'm interested in picking this up over the weekend if that is of interest, alternatively using human-panic looks like it would be straightforward enough. Do you have any specific requirements about what kind of information you would like to include in the panic message?

Profpatsch:
After some deliberation (and looking at the `human_panic` source), I’ve arrived at the conclusion that we should use the `human_panic` crate, and create a few issues/PRs to improve it.

The only dependencies that are a bit questionable are `termcolor` and `failure`.
`termcolor` is only used to color the output red, which I’d argue is not really worth a dependency that wraps terminal behavior.
`failure` is a bit strange, especially because I didn’t see the code use it in any way.

So let’s just add `human_panic` for now and add a tracking issue for removing these deps. We already depend on `termcolor` via `env_logger` anyway.

johnchildren:
Thanks! I have made a PR to human-panic https://github.com/rust-cli/human-panic/pull/63 which should update a few things and (if accepted) it seems that the dependency on `failure` was unnecessary as well as pruning a few other crates that weren't being used.

Unfortunately the version of human-panic in crates.io has not yet been updated for a version of termcolor >= 1.0 so two versions of the crate would be introduced if I were to add the current version of the crate, but potentially could be done for now and then updated when a new version is released?

Profpatsch:
Great work!



> Unfortunately the version of human-panic in crates.io has not yet been updated for a version of termcolor >= 1.0 so two versions of the crate would be introduced if I were to add the current version of the crate, but potentially could be done for now and then updated when a new version is released?

In the optimal case `human_panic` would have a config option that disables term colors (I don’t think they are needed to achieve the libraries goal, and coloring terminal output for errors usually interferes with users setup anyway). But if they definitely want to keep those, it shouldn’t be that much work to update to >1.0, right? It’s just three lines in the code, which sets the buffer color to red.

[...]

PR #181, opened 2019-10-20
Add human-panic

johnchildren:
Add human-panic crate which provides a handler for catching panics and
providing a more human friendly message including contact information
and where to submit bugs as directed by Cargo.toml.

Current provided as pinned to a git commit pending a 2.0 release of
human-panic.

NOTE: In order to get the panic message to appear the binary must both
be a release target and the RUST_BACKTRACE environment variable must be
unset for it to appear. Setting RUST_BACKTRACE to 0 or similar will give
you a regular panic without a backtrace but will not display this
message.

Example panic output from panicking immediately after setup:

```
Well, this is embarrassing.

lorri had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/run/user/1000/report-eb81bb42-18a3-4be5-8e45-3143269d8274.toml". Submit an issue or email with the subject of "lorri Crash Report" and include the report as an attachment.

- Homepage: https://github.com/target/lorri
- Authors: Graham Christensen <graham.christensen@target.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```

Example file content:

```
name = 'lorri'
operating_system = 'unix:Unknown'
crate_version = '0.1.0'
explanation = '''
Cause: explicit panic. Panic occurred in file 'src/main.rs' at line 23
'''
cause = 'Error cause could not be determined by the application.'
method = 'Panic'
backtrace = '''

   0: 0x5611a588540a - __rust_maybe_catch_panic
                at src/libpanic_unwind/lib.rs:87
   1: 0x5611a587ef6d - try<i32,closure>
                at src/libstd/panicking.rs:272
                 - catch_unwind<closure,i32>
                at src/libstd/panic.rs:388
                 - lang_start_internal
                at src/libstd/rt.rs:48
   2: 0x5611a57b7cf2 - main
   3: 0x7f7c6bf79b8e - __libc_start_main
   4: 0x5611a57b441a - _start
   5:        0x0 - <unknown>'''
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 authors = [
   "Graham Christensen <graham.christensen@target.com>",
 ]
+homepage = "https://github.com/target/lorri"
 license = "Apache-2.0"
 
 [dependencies]
@@ -25,3 +26,6 @@ atomicwrites = "0.2.3"
 vec1 = "1.1.0"
 proptest = "0.9.1"
 nix = "0.14.0"
+# pinned to a git commit pending v2.0 release
+# see: https://github.com/rust-cli/human-panic/issues/46
+human-panic = { git = "https://github.com/rust-cli/human-panic.git", rev = "70db9489" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,26 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "backtrace"
+version = "0.3.38"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace-sys 0.1.31 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.55 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-demangle 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace-sys"
+version = "0.1.31"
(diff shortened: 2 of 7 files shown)


johnchildren:
Should resolve: https://github.com/target/lorri/issues/114

Cargo.toml
```diff
@@ -24,4 +25,5 @@ tempfile = "3.0.7"
 atomicwrites = "0.2.3"
 vec1 = "1.1.0"
 proptest = "0.9.1"
-nix = "0.14.0"
\ No newline at end of file
+nix = "0.14.0"
+human-panic = { git = "https://github.com/rust-cli/human-panic.git", rev = "70db9489" }
```
johnchildren: This could just be set to master and then enforced by the lockfile, am unsure what is preferred.

I should probably add a comment as well.

johnchildren:
Ah forgot to regenerate Cargo.nix, will try to do so tomorrow.

johnchildren:
Should be passing now, had to add nix-prefetch-git to the shell environment.

Profpatsch:
It’s missing the `homepage` field, I think because carnix doesn’t generate it.

I [have a PR for switching to naersk](https://github.com/target/lorri/pull/193), which means patching carnix now might not be worth the effort. Depending on how fast I can get that to work, I’ll merge that first and rebase this PR onto it.

johnchildren:
Unfortunately I think Naersk doesn't support git dependencies (from the [README](https://github.com/nmattia/naersk/blob/master/README.md#comparison)), so we may need to try and get this working with a released version instead.

Profpatsch:
> Unfortunately I think Naersk doesn't support git dependencies

It should be able to patch naersk slightly, so that we can override with a specific nix version.

Profpatsch:
Rebased on master and added the homepage field when built from `default.nix`.

Profpatsch:
Great, it should work now.

Issue #145, opened 2019-09-04
Indicate that Lorri is done building for a project


nyarly:
With the conversion to a single daemon, it becomes difficult to tell whether a particular project is done building, which is important for knowing what kinds of things are "safe" to do. It's reasonable to update documentation while building is happening, but not run tests, for instance.

Three separate things would be helpful for this UI:
* A status bar indicator would be helpful in this case. Such an indicator will tend to be current when a tracked file changes (i.e. it will warn in time) but lag indicating "all clear" - the user needs to "poll" it by triggering a new status update.
* A libnotify/growl/whatever notification that build is complete would cover the lag. `notify-rust` is a nice crate for this feature.
* A `is-building` subcommand for use in build scripts and tools would be helpful: having a large test suite exit early if lorri is updating dependencies would be far better than it failing late with an ambiguous error.

nyarly:
Honestly, `lorri is-building` used in a prompt definition probably satisfies the first case.

It may be better to have something like an event stream socket or something for the second, and let users set up their own notifications if they'd like. Weakest level shill: https://crates.io/crates/socket-notify could be adapted to serve that purpose if needed.

Profpatsch:
Would a plumbing command like `lorri daemon stream-events_` satisfy this?

```
$ lorri daemon stream-events_
{ "file": "/my/project/shell.nix", "event": "started", "data": {} }
{ "file": "/my/project/shell.nix", "event": "failed", "data": { … interesting data … }  }
```

Then stuff like
```
for l in $(lorri daemon stream-events_); do
    if $(echo "$l" | jq --exit-status '.file == "/my/project/shell.nix"'); then
        echo "$l" | jq --raw-output '.event' | xargs -i {} notify-send {}
    fi
done
```
can be used as easy integration into status bars / scripts.

If we want to always print the current status (as opposed to just new events as they come in) first, it gets a tad more complicated to implement, but should still be doable.

roberth:
Seems to overlap #79

Issue #79, opened 2019-05-13
Feature Request: Evaluation Notifications


gilligan:
**in short**:
As a lorri user I would like to be able to be notified whenever an evaluation has finished and know whether it was successful or not.

**current scenario**:
I am working on a Haskell project and add a dependency to my cabal file. In order to know if this was successful and when I can start using it I have to open my my (otherwise hidden because mostly uninteresting) tmux pane with the lorri process.

**desired scenario**
I am working on a Haskell project and add a dependency to my cabal file. Everything is built and re-evaluated successfully and lorri triggers a notification which I can see through my dbus notification app.

-----
**Some additional thoughts**
- I understand lorri will eventually be running as a daemon in the background. In this case support for a notification like this might be even more desirable. 
- I think #64 is related to this and it would in fact be possible to use this generic mechanism to trigger notifications using something like `notify-end` yet I believe that being informed about the state of the evaluation is important enough to be a feature in its own right.
- Instead of only notifying about successful or unsuccessful evaluation completion it might also be useful to include evaluation start. Ideally this would be configurable on the command line.
- https://github.com/hoodie/notify-rust

Looking forward to hear your thoughts on this. Thanks

Profpatsch:
We could add a `--notify` flag to the tool that watches for changes.

Would you expect to be notified about every evaluation on every project the daemon watches, or just for some subset of projects?

gilligan:
@Profpatsch Oh that's a good question. I think I would say _every_ project because everything else would probably be inconvenient and unexpected. Plus evaluations should only occur if I actually *do* change something in the project and then i would also want to be informed about it.

domenkozar:
It seems this feature has some intersection with #64 to emit events

Issue #78, opened 2019-05-09
Expose build event stream


Profpatsch:
In order to support use-cases like https://github.com/target/lorri/issues/64 and make lorri more of a toolbox, we should provide a way for users to hook into the stream of build events the daemon generates.

Right now, we print the corresponding events to stderr:

https://github.com/target/lorri/blob/71983296d121a8bc1af7dbeec78b7aa6dacf1e46/src/build_loop.rs#L15-L23

A command like `lorri listen-for build-events` could be added, which produces a stream of `\0`-separated JSON objects. (If \0 might occur in the JSON, netstrings are an alternative.)  The exact contents of these objects depend on the potential use-cases for such a command, but should contain at least the nix file that was built and whether the build succeeded.

Shados:
JSON values aren't allowed to contain _unescaped_ newlines, so you can just print your objects without whitespace and do one object per line :). No need for null-separated objects, and this approach meshes well with most stdin-consuming tools. It will also work fine with `jq`'s default of per-streamed-object filter application.

domenkozar:
I'd highly recommend using http://jsonlines.org/

PR #183, opened 2019-10-21
Event streaming

nyarly:
Adds a `lorri stream_events_` subcommand that receives json formatted representations of events. Each event is its own `\n` separated line. The command takes an option to toggle current state reporting vs. live events (or both), which is useful for the motivating use cases.

Included are example jq|xargs style scripts to make use of the event streams.


t184256:
Just a user report: I've tried that and it worked fine for me.

nyarly:
Closes #145 #78 #79 (I believe)

src/build_loop.rs
```diff
@@ -9,32 +9,157 @@ use crate::project::roots;
...
+#[serde(rename_all = "snake_case")]
 pub enum Event {
-    /// The build has started
-    Started(Reason),
-    /// The build completed successfully
-    Completed(BuildResults),
-    /// The build command returned a failing exit status
-    Failure(BuildExitFailure),
+    /// A heartbeat to send over sockets to keep them alive
+    Heartbeat,
```
Profpatsch: Design question: Do we absolutely *need* to send a heartbeat? I’d like to keep the state machine as simple as possible. How about we just try to send a message and throw the client out if we can’t send it?

In that case, we would just stop the client command if the connection is broken for whatever reason (let’s assume the socket can connect always if the daemon is running). No fallbacks, no retries, no heartbeats.

nyarly: The client (all socket::Readers iiuc) reads on a timeout. Since there's an arbitrary amount of time between builds, no length of timeout is going to be sufficient.

It could be that once the client times out (and uses a much longer timeout...), it attempts to reconnect. I'm not sure what the consequences there would be, apart from constant re-opens of sockets. That might well be perfectly benign; I'm not sure.

Profpatsch: `Timeoute::Infinite` should do what you want, together with `read()` on a `ReadWriter`.

nyarly:
I'll rebase and squash to a single commit when I get a chance.

src/build_loop.rs
nyarly: I'll give it a whirl. If it's appropriate here, why not for ping, though?

nyarly:
Merged with master, removed Heartbeats.

nyarly:
@curiousleo I'm trying to figure out how varlink interacts here. How can I get the client to exit if the server takes to long to respond after a connect?

That seems to happen when the worker threads max has been hit.

It also seems like, having set up a more/continues stream of replies, that the worker threads won't expire except by idle timeout. That includes when the client quits and the socket connection closes.

To be clear: I'm still working on proving this diagnosis; pointers to varlink diagnostics would be helpful as well.

curiousleo:
> @curiousleo I'm trying to figure out how varlink interacts here. How can I get the client to exit if the server takes to long to respond after a connect?

I'm looking at this line for context: https://github.com/target/lorri/pull/183/files#diff-d088683aab00cb5a8409ddee9949c956R56. As far as I can tell, the client API does not expose any timeout-related options. So without async streams (*), the only thing I can think of is to run the actual client in one thread and have a separate supervisor thread which terminates the client after a certain amount of time if there has been no activity. The two threads could communicate via atomics, for example.

(*) With async Streams, you could just merge the `client.monitor().more()` stream with a stream that emits a single "timeout" event after a certain amount of time.

> That seems to happen when the worker threads max has been hit.

For context, I'm looking here: https://github.com/target/lorri/pull/183/files#diff-cd0ba2c39443e5321251948314aabea8R52. I wonder if perhaps a pragmatic solution for this for the time being could be to set the maximum number of server threads to "effectively infinite", i.e. some very high number. The library only creates as many worker threads as required, rather than always spawning the maximum number of worker threads, so that should be fine ... i.e. as a user you'll just have to accept that if you have a very large number of projects, lorri is going to need some resources to handle them.

> It also seems like, having set up a more/continues stream of replies, that the worker threads won't expire except by idle timeout. That includes when the client quits and the socket connection closes.

Interesting. I wonder if this is due to some restriction of the underlying stream API or a bug.

> To be clear: I'm still working on proving this diagnosis; pointers to varlink diagnostics would be helpful as well.

Cool! Looking around the varlink API, I couldn't find much that looked like it would provide access to diagnostics. What kinds of numbers were you looking for? My guess is the number of "busy" server-side threads?

nyarly:
Following up: it appears that worker threads block in `call.reply(event)` - debugging output emits the event just before, and a `dbg` macro around the `reply` doesn't print.

Separately, `netstat -x` prints an ever increasing number of connections to the lorri socket, even though the clients are ephemeral.

Forcing Lorri to send messages by `touch shell.nix`ing emits broken pipe errors from those `dbg`s

@Profpatsch I'm starting more an more to think that the only way to manage this is periodic heartbeats over the socket to trigger varlink noticing that the socket closed. Together with @curiousleo's recommendation that we extend the worker thread count, I think that'd lead to reasonable behavior.

Without some mechanism to keep sockets closing, we'll still sometimes hit whatever headroom (artificial or otherwise) we have on thread/socket counts between build runs.

Maybe the NewListener event makes sense to send to all the consumers? The clients could drop it unprinted, but it'd give the workers something to exit on if their client was gone?

nyarly:
As a workaround, I've implemented a server deadline in `lorri stream_events_` and echo NewListener to all existing clients.

Profpatsch:
> @Profpatsch I'm starting more an more to think that the only way to manage this is periodic heartbeats over the socket to trigger varlink noticing that the socket closed.

yeah, if that’s the case go for it. My request for removal came from before the decision was made to throw the socket code away and switch to varlink. Sorry for the overhead.

src/daemon.rs
```diff
@@ -39,21 +55,26 @@ pub struct Daemon {
...
-        let (build_tx, build_rx) = chan::unbounded();
+    pub fn new() -> (Daemon, chan::Receiver<LoopHandlerEvent>) {
+        let (tx, rx) = chan::unbounded();
+        let (mon_tx, mon_rx) = chan::unbounded();
         (
             Daemon {
                 handler_threads: HashMap::new(),
-                build_tx,
+                build_events_tx: tx,
+                build_events_rx: rx.clone(),
```
curiousleo: There should be no need to clone `rx` here. I would suggest renaming `tx` to `build_events_tx` and `rx` to `build_events_rx`, that way you can use the shorthand struct instantiation notation for `Daemon`, e.g. `Daemon { handler_threads: HashMap::new(), build_events_tx, build_events_rx, mon_tx } `.

src/build_loop.rs
```diff
@@ -57,8 +179,11 @@ impl<'a> BuildLoop<'a> {
     /// When new filesystem changes are detected while a build is
     /// still running, it is finished first before starting a new build.
     #[allow(clippy::drop_copy, clippy::zero_ptr)] // triggered by `select!`
-    pub fn forever(&mut self, tx: chan::Sender<Event>, rx_ping: chan::Receiver<()>) {
-        let send = |msg| tx.send(msg).expect("Failed to send an event");
+    pub fn forever(&mut self, tx: chan::Sender<LoopHandlerEvent>, rx_ping: chan::Receiver<()>) {
+        let send = |msg| {
+            debug!("BuildLoop sending"; "message" => format!("{:#?}", msg));
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/com.target.lorri.varlink
```diff
@@ -12,6 +12,38 @@ type ShellNix (
...
 )
 
+# Monitor the daemon. The method will reply with an update whenever a build begins or ends.
+# Montior will immediately reply with a snapshot of known projects, then a marker event,
+# indicating that the stream of events is now "live."
+method Monitor() -> (event: Event)
+
+type Event (
+    kind: (sectionend, started, completed, failure),
+    nix_file: ?ShellNix, # not in sectionend
```
curiousleo:
```suggestion
    nix_file: ?ShellNix, # not present if kind == sectionend
```
That would be my suggestion - it's just a little more explicit. I would do something similar for the other comments.

src/com.target.lorri.varlink
```diff
@@ -12,6 +12,38 @@ type ShellNix (
...
+    reason: ?Reason,     # only if started
+    result: ?Outcome,     # only if completed
```
curiousleo: Please align comment :)

src/com.target.lorri.varlink
```diff
@@ -12,6 +12,38 @@ type ShellNix (
   path: string
 )
 
+# Monitor the daemon. The method will reply with an update whenever a build begins or ends.
+# Montior will immediately reply with a snapshot of known projects, then a marker event,
+# indicating that the stream of events is now "live."
+method Monitor() -> (event: Event)
+
+type Event (
+    kind: (sectionend, started, completed, failure),
```
curiousleo:
```suggestion
    kind: (section_end, started, completed, failure),
```

src/com.target.lorri.varlink
```diff
@@ -12,6 +12,38 @@ type ShellNix (
...
+    nix_file: ?ShellNix, # not in sectionend
+    reason: ?Reason,     # only if started
+    result: ?Outcome,     # only if completed
+    failure: ?Failure    # only if failure
+)
+
+type Reason (
+    kind: (projectadded, pingreceived, fileschanged, unknown),
```
curiousleo:
```suggestion
    kind: (project_added, ping_received, files_changed, unknown),
```

src/daemon.rs
```diff
@@ -65,13 +86,59 @@ impl Daemon {
...
                 .expect("failed to serve daemon server endpoint");
         })?;
+        let build_events_rx = self.build_events_rx.clone();
+        let mon_tx = self.mon_tx.clone();
+        pool.spawn("build-loop", move || {
+            let mut project_states: HashMap<NixFile, Event> = HashMap::new();
+            let mut event_listeners: Vec<chan::Sender<Event>> = Vec::new();
+
+            for msg in build_events_rx {
+                debug!("Build loop message"; "message" => ?msg);
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/daemon.rs
```diff
@@ -65,13 +86,59 @@ impl Daemon {
...
+                match &msg {
+                    LoopHandlerEvent::BuildEvent(ev) => match ev {
+                        Event::SectionEnd => (),
+                        Event::Started { nix_file, .. }
+                        | Event::Completed { nix_file, .. }
+                        | Event::Failure { nix_file, .. } => {
+                            project_states.insert(nix_file.clone(), ev.clone());
+                            event_listeners.retain(|tx| {
+                                let keep = tx.send(ev.clone()).is_ok();
+                                debug!("Sending"; "event" => format!("{:#?}", ev), "keep" => keep);
```
curiousleo:
```suggestion
                                debug!("Sent"; "event" => ?ev, "keep" => keep);
```

See https://docs.rs/slog/2.5.2/slog/macro.log.html#fmtdisplay-and-fmtdebug-values.

src/daemon.rs
```diff
@@ -65,13 +86,59 @@ impl Daemon {
...
+                                keep
+                            })
+                        }
+                    },
+                    LoopHandlerEvent::NewListener(tx) => {
+                        debug!("adding listener");
+                        let keep = project_states.values().all(|event| {
+                            let keeping = tx.send(event.clone()).is_ok();
+                            debug!("Sending snapshot"; "event" => format!("{:#?}", &event), "ok" => keeping);
```
curiousleo:
```suggestion
                            debug!("Sent snapshot"; "event" => ?&event, "keep" => keeping);
```

src/daemon.rs
```diff
@@ -65,13 +86,59 @@ impl Daemon {
...
+                            keeping
+                        });
+                        debug!("Finished snapshot"; "keep" => keep);
+                        if keep {
+                            event_listeners.push(tx.clone());
+                        }
+                        event_listeners.retain(|tx| {
+                            let keep = tx.send(Event::SectionEnd).is_ok();
+                            debug!("Sending new listener sectionend"; "keep" => keep);
```
curiousleo:
```suggestion
                            debug!("Sent new listener sectionend"; "keep" => keep);
```

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
             Err(e) => call.reply_invalid_parameter(e),
         }
     }
+
+    fn monitor(&self, call: &mut dyn rpc::Call_Monitor) -> varlink::Result<()> {
+        debug!("starting monitor worker loop");
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
...
+                    debug!("translated"; "varlink event" => format!("{:#?}", &ev));
+                    call.reply(ev)
+                }
+                Err(e) => {
+                    debug!("receiving event"; "error" => format!("{:#?}", &e));
+                    call.reply_invalid_parameter(e.to_string())
+                }
+            }?;
+        }
+        debug!("worker loop done for monitor");
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
...
+        }
+
+        let (tx, rx) = chan::unbounded();
+        self.build_tx
+            .send(LoopHandlerEvent::NewListener(tx))
+            .map_err(|_| varlink::error::ErrorKind::Server)?;
+
+        call.set_continues(true);
+        for event in rx {
+            debug!("event for varlink"; "event" => format!("{:#?}", &event));
```
curiousleo:
```suggestion
            debug!("Monitor event for varlink"; "event" => ?&event);
```

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
...
+            match event.try_into() {
+                Ok(ev) => {
+                    debug!("translated"; "varlink event" => format!("{:#?}", &ev));
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
...
+                    call.reply(ev)
+                }
+                Err(e) => {
+                    debug!("receiving event"; "error" => format!("{:#?}", &e));
```
curiousleo: This debug message seems likely to cause spam. It's probably useful for building this feature, but I suggest removing it before this PR is merged.

src/daemon/rpc.rs
```diff
@@ -73,9 +81,307 @@ impl rpc::VarlinkInterface for Server {
...
+            log_lines: rf
+                .log
+                .into_iter()
+                .map(|l| {
+                    let oss: OsString = l.into();
+                    build_loop::LogLine::from(oss)
+                })
+                .collect(),
+        }
+        */
```
curiousleo: Is this meant to be removed?

src/ops/stream_events.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+#[derive(Debug)]
+enum Error {
+    Varlink(rpc::Error),
+    Compat(String),
+}
+
+/// See the documentation for lorri::cli::Command::Shell for more
```
curiousleo: Hm ... :)

src/ops/stream_events.rs
```diff
@@ -0,0 +1,101 @@
...
+/// details.
+pub fn main(kind: EventKind) -> OpResult {
+    // TODO: set up socket path, make it settable by the user
```
curiousleo: You are getting the socket path here, right? I don't understand this TODO.

curiousleo: You are getting the socket path here, right? I don't understand this TODO.

src/ops/stream_events.rs
```diff
@@ -0,0 +1,101 @@
...
+    debug!("Starting stream_events");
```
curiousleo: This seems unnecessary. Nothing has happened yet, so this debug line doesn't really contain any information.

src/ops/stream_events.rs
```diff
@@ -0,0 +1,101 @@
...
+    let address = get_paths()?.daemon_socket_address();
+
+    use rpc::VarlinkClientInterface;
+    let mut client = rpc::VarlinkClient::new(
+        varlink::Connection::with_address(&address).expect("failed to connect to daemon server"),
+    );
+
+    let mut snapshot_done = false;
+    let (tx, rx) = unbounded();
+    let recycle = tx.clone();
```
curiousleo: `tx_clone` perhaps? `recycle` as a name doesn't carry much intuition, at least for me.

src/ops/stream_events.rs
```diff
@@ -0,0 +1,101 @@
...
+            Ok(ev) => match (snapshot_done, &kind) {
+                (_, EventKind::All) | (false, EventKind::Snapshot) | (true, EventKind::Live) => {
+                    println!(
+                        "{}",
+                        serde_json::to_string(&ev).expect("couldn't serialize event")
+                    )
+                }
+                _ => (),
+            },
+            Err(err) => return Err(ExitError::temporary(format!("{:?}", err))),
```
curiousleo: This could be moved into the `match` expression, i.e. appended to line 76. Something like:

```rust
         match event
             .map_err(Error::Varlink) 
             .and_then(|e| e.try_into().map_err(Error::Compat)) 
             .map_err(ExitError::temporary(format!("{:?}", err)))?
         {
```

Then you only need to handle the `Ok` case inside the `match`.

src/project/roots.rs
```diff
@@ -19,7 +19,7 @@ pub struct Roots {
 
 /// A path to a gc root.
 #[derive(Hash, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
-pub struct RootPath(PathBuf);
+pub struct RootPath(pub PathBuf);
```
curiousleo: Not super elegant IMO, but I think `RootPath` should just go away at some point, so I don't mind breaking the abstraction here.

example/notify.sh
```diff
@@ -0,0 +1,9 @@
+#! /usr/bin/env nix-shell
```
curiousleo: We have a `contrib` directory in the repo already for "extra stuff around lorri that could be useful", like `notify.sh` and `prompt.sh`.

I think these scripts are really great in showing what users can do with this feature. Would you mind
- moving the scripts to `contrib`
- adding a tiny little guide in the style of the ones already present in `contrib` (see https://github.com/target/lorri/tree/master/contrib)?

curiousleo: If you prefer, you can also remove `notify.sh` and `prompt.sh` from this PR and re-introduce them with docs in a follow-up PR. That could help keep the size of this PR down a bit. Whichever you think is easier.

src/cli.rs
```diff
@@ -122,6 +127,14 @@ pub struct Ping_ {
     pub nix_file: PathBuf,
 }
 
+/// Stream events from the daemon.
+#[derive(StructOpt, Debug)]
+pub struct StreamEvents_ {
+    #[structopt(short, long, default_value = "all")]
```
curiousleo: My suggestion would be to leave out the short option here. If we really think an additional short option would improve UX, we can always add it afterwards. On the other hand, removing it at some later point is a [breaking change](https://github.com/target/lorri/blob/master/MAINTAINERS.md#major-or-minor-release). So my suggestion would be to keep the user interface minimal.

src/com.target.lorri.varlink
```diff
@@ -12,6 +12,38 @@ type ShellNix (
   path: string
 )
 
+# Monitor the daemon. The method will reply with an update whenever a build begins or ends.
+# Montior will immediately reply with a snapshot of known projects, then a marker event,
+# indicating that the stream of events is now "live."
+method Monitor() -> (event: Event)
+
+type Event (
```
curiousleo: Varlink does not provide proper enums or gRPC-like `oneof`. But I think that events can be modelled in a slightly more "type-safe" way. What do you think about this sort of structure:

```
type Event {
    kind: (section_end, started),
    section_end: SectionEnd?, # set iff kind == section_end
    started: Started?, # set iff kind == started
    # ...
}
```

where `SectionEnd`, `Started`, etc. have only non-optional fields?

nyarly:
@curiousleo Thank you for the comprehensive and thoughtful review.

Some of your suggestions are already addressed in my local branch. I'll be pushing those changes presently.

Of the remainder, I like all of them and intend to implement.

I am concern is that the pace of development outstrips my available time. This week has been merging current master back to this branch. If we could merge a sub-optimal version, I'm happy to commit to followup PRs.

src/ops/stream_events.rs
nyarly: This and the doc comment were copypasta :/

src/ops/stream_events.rs
nyarly: I was hoping to indicate the reason for this channel - to feed a first event back into our local channel that's tee'd from the daemon monitor.

example/notify.sh
nyarly: Done. That'll be first follow-up, since it also makes to elaborate the daemon.md with notify.sh

src/cli.rs
nyarly: That's a good point. I was cribbing from elsewhere, but the plumbing commands don't need to have shortcuts like that.

src/com.target.lorri.varlink
nyarly: I think this might be interesting to experiment, but I think I need to put it into a followup PR.

src/ops/stream_events.rs
curiousleo: That's fine. I find naming channels particularly difficult - I find myself looking at their uses a lot to figure out what they're actually meant to do. You _could_ document this channel's purpose in a comment if you think that would be useful and possible to do succinctly. If not, don't - your call :)

src/ops/stream_events.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    });
+
+    select! {
+        recv(rx) -> event => recycle.send(event.expect("local channel couldn't receive")).expect("local channel couldn't resend"),
+        default(Duration::from_millis(250)) => return Err(ExitError::temporary("server timeout"))
+    }
+
+    for event in rx.iter() {
+        debug!("Received"; "event" => format!("{:#?}", &event));
```
curiousleo: One last `format!` that can be replaced by slog's `?&event` syntax.

tests/daemon/main.rs
```diff
@@ -17,7 +17,7 @@ use std::time::{Duration, Instant};
 pub fn start_job_with_ping() -> std::io::Result<()> {
     // TODO: this code is a mess because Daeomon is not
     // nicely abstracted yet.
-
+    //
```
curiousleo: Super-nit: this line change looks superfluous.

curiousleo:
@nyarly thanks for the work you put in here, and for being very responsive to code review.

> I am concern is that the pace of development outstrips my available time. This week has been merging current master back to this branch.

That's totally fair. The last couple of weeks were particularly frenetic. Expect fewer sweeping changes from now on.

> If we could merge a sub-optimal version, I'm happy to commit to followup PRs.

I'm on board with that idea, with one _big caveat_ I'll expand on in my next comment.

The two items of I'd like to see follow-up on are:

- https://github.com/target/lorri/pull/183#discussion_r373391813: a PR that brings back `notify.sh` and `prompt.sh` and explains how users can leverage event streams
- https://github.com/target/lorri/pull/183#discussion_r373508954: a PR that makes it harder to use the event streams varlink protocol the wrong way by making it more type-safe (note that the varlink protocol is considered an implementation details, so changes to it are not considered breaking changes even after a 1.0 release)

I think that you have the most context on this and are probably the best person to do these things, but if you'd like help or would like to work on them together, we can definitely talk about that.

curiousleo:
#### Problem: implementation details exposed in the output

I think there is an issue with re-using existing structs the way it is done right now in this PR. Here's an example:

```console
$ lorri internal__stream_events | jq
{
  "started": {
    "nix_file": {
      "shell": "/home/leo/Code/lorri/shell.nix"
    },
    "reason": {
      "project_added": {
        "shell": "/home/leo/Code/lorri/shell.nix"
      }
    }
  }
}
{
  "completed": {
    "nix_file": {
      "shell": "/home/leo/Code/lorri/shell.nix"
    },
    "result": {
      "output_paths": {
        "shell_gc_root": "/home/leo/.cache/lorri/gc_roots/2043910c8519bf079838568f54a73839/gc_root/shell_gc_root"
      }
    }
  }
}
```

What is clear here is that this exposes
https://github.com/target/lorri/blob/c38e0f09c36a1b27d86eb01ba5672fdffa950641/src/lib.rs#L50-L55 https://github.com/target/lorri/blob/c38e0f09c36a1b27d86eb01ba5672fdffa950641/src/watch.rs#L34-L44 https://github.com/target/lorri/blob/c38e0f09c36a1b27d86eb01ba5672fdffa950641/src/builder.rs#L284-L287 and other structs and enums to users of this command - basically all those that now have `Serialize` derives.

This means that any change to any of those structs or enums is likely to change the output format of `event_stream`. I think it is likely that we'll want the freedom to change those structs in the future. I also think that the way things work at the moment in this PR, it is easy to make a change to any of them and not realise that this changes the format of `event_stream`'s output.

#### Context: v1.0 and backwards compatibility

We are working towards a lorri v1.0 release: https://github.com/target/lorri/issues/306. From that point on, the [versioning scheme](https://github.com/target/lorri/blob/master/MAINTAINERS.md#versioning-scheme) will be used to determine the next version.

I believe that @Profpatsch may have planned for `event_stream` to be a non-internal command.

_If_ `event_stream` is to be labelled non-internal, then _after_ we release v1.0, any change to any of the structs or enums that can appear in `event_stream`'s output will be a "change to the machine-readable output of a non-internal command", which means it's a breaking change, which means it requires a major version bump. That's not the end of the world, but the way things stand right now, I think that any internal change is likely to touch at least one of the enums and structs exposed via `event_stream`, so my fear is that we'd race through lots of major versions very quickly.

#### Proposal: expose an explicit interface

My proposal is thus to separate the internal representation of things like build failure reasons from how they are represented in the output of the `event_stream` command. This allows us to make changes to the internal representations without changing the output format. It also makes it harder to accidentally change the output format of `event_stream`.

I see two options for how the explicit interface can be defined:

- We can write a set of Rust structs and enums that serve the particular purpose of being serialised into the output format for `event_stream`.
- We can use Varlink type definitions to describe the output format. Describing a JSON schema is something Varlink is pretty good at, in my opinion. People who want to build tooling for lorri using `event_stream` may find it easier to read the Varlink schema than Rust structs and enums. Finally, changes to this definition would be very obvious, this would pretty much rule out making output format changes by accident. The downside is that we would now have two sets of Varlink definitions: an internal one for the daemon <> lorri clients communication, and this new one describing the `event_stream` output. They should probably go in separate files and be documented appropriately to make sure future collaborators understand their different purposes.

---

I hope this all makes sense. To summarise: `event_stream` exposes types that I think of as implementation details. As a result, we'll either change the output of `event_stream` a lot (perhaps unwittingly), or we'll slow down development because we try not to change the output of `event_stream`, which means preserving many types as they are. Neither is a good place to be, in my opinion, even ignoring the versioning issue. I think the way to get around this problem is to separate the representation of the user-facing output from the internal representation.

curiousleo:
https://github.com/target/lorri/pull/320#issuecomment-581421805 describes my proposal for our next steps with this work.

PR #184, opened 2019-10-22
Error type refactoring

nyarly:
Some small changes with OpResult that result in slightly less ceremony around error handling in the ops module.

In part, this was an exercise in type manipulation in Rust, but I think the results are pleasing in a small way.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,7 @@ use lorri::locate_file;
 use lorri::NixFile;
 
 use lorri::cli::{Arguments, Command};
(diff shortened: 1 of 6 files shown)


Profpatsch:
Thanks for this, when looking at the `ExitError` code as it stands now, I noticed that we are leaking the internals to the users.

So instead of exposing the error code, I opted for making the struct entirely private and only going through constructor methods, in https://github.com/target/lorri/pull/199.

I think I included most of your changes here, except for the `err` helper function. I credited you in the first commit message.

Issue #85, opened 2019-05-27
Display warnings if certain well-known directories (e.g. .git) are imported 


Profpatsch:
Lorri should make an effort to help the user prevent common mistakes, by displaying (silence-able?) warnings.

* `.git` directories
* Directories that are very large

In another step, if we know what kind of project the user has, we can check for language-specific directories (e.g. `target/` for Rust/`cargo` projects). Or even implement that logic for them with our own library of `mkShell`-like functions (`lorriRustProject`, `lorriGoProject` …).

Profpatsch:
A first step is to collect a series of directories which should be linted.

An idea is to use `.gitignore` to find ignored files and print a warning if one of those is added to the watcher.

New files are currently added here:

https://github.com/target/lorri/blob/master/src/ops/daemon.rs#L60-L72

so that’s the place where the linter should hook into.

Jeschli:
I'll take that one.

Jeschli:
So I did some code onion peeling today. Where do you want a linter to live?
I would suggest to extend the Project struct with some linter object. So in buildloop.once a 
linter.do_warnings(&paths) can be done.

Does this make sense? If you guys agree, or want to see it. I'll prepare a WIP PR.

Cheers

PR #187, opened 2019-10-23
WIP: build loop: first try of a linter #85

Jeschli:


curiousleo:
(Relates to https://github.com/target/lorri/issues/85)

src/build_loop.rs
```diff
@@ -96,6 +99,7 @@ impl<'a> BuildLoop<'a> {
         let (tx, rx) = channel();
         let run_result = builder::run(tx, &self.project.nix_file, &self.project.cas)?;
 
+        self.linter.check(&run_result.referenced_paths);
```
Profpatsch: This is a good start, though it matches a bit too much.

If you look at

https://github.com/target/lorri/blob/8c5a3fbd0dbeb6d54b6a9ea3eac10ea03e0f770f/src/builder.rs#L154-L184

You see that `paths` (which turns into `referenced_paths` at the end) is constructed from multiple `LogDatum` sources

https://github.com/target/lorri/blob/8c5a3fbd0dbeb6d54b6a9ea3eac10ea03e0f770f/src/builder.rs#L298-L311

We only really need the `CopiedSource` case, because that’s where the user can make a mistake (in copying to much because of a wrong `builtins.filterSource` filter).

So we should add a new field to the return of `instrumented_instantiation`, which returns the copied sources and then pipe that through `builder::run()` and out to `build_loop`.

I can help with that if that’s too much overhead.

PR #188, opened 2019-10-23
Pin nixpkgs to current NixOS 19.09 snapshot

curiousleo:
The current pinned snapshot is five months old. Updating it means fewer things need to be downloaded initially.

### DIFF
--- a/nix/nixpkgs.json
+++ b/nix/nixpkgs.json
@@ -1,7 +1,7 @@
 {
   "url": "https://github.com/nixos/nixpkgs-channels.git",
-  "rev": "c7bcd4277cc9a656207b636dbc62fef21dc64c78",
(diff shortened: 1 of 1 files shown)


curiousleo:
CI failure: https://travis-ci.org/target/lorri/jobs/601773976#L2229

```
running 1 test

test tests::trivial_shell_nix ... FAILED

failures:

---- tests::trivial_shell_nix stdout ----

thread 'tests::trivial_shell_nix' panicked at 'stdout:

stderr:error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I), at (string):2:17

', src/main.rs:100:9
```

I'm not sure in what way this relates to my change.

curiousleo:
PTAL

PR #189, opened 2019-10-23
[WIP] lorri daemon --once

curiousleo:


Profpatsch:
Hm, we already have `lorri watch --once` I don’t think `lorri daemon --once` would give us anything. Plus it’s not at all clear to me what that would do in the first place.

curiousleo:
> Hm, we already have `lorri watch --once` I don’t think `lorri daemon --once` would give us anything. Plus it’s not at all clear to me what that would do in the first place.

My understanding was that `lorri watch` is to be phased out. This is an attempt to move towards feature parity between `lorri daemon` and `lorri watch`. As I understood @grahamc, `lorri watch --once` was introduced mainly for testing / debugging purposes. `lorri daemon --once` would have the same use case.

If you think that `lorri watch --once` isn't really necessary either, great! :) Then there's no need to port it.

curiousleo:
Relevant: https://github.com/target/lorri/pull/161

curiousleo:
As I understand https://github.com/target/lorri/pull/161#issue-323906585, the idea of `lorri run` is to make the _client_ block on and exit after a single build. My experiment in this PR was to make the _daemon_ block on and exit after a single build.

I believe that a blocking build-then-exit _client_ command is more useful, so I think we're in agreement here @Profpatsch. I'll close this PR.

PR #191, opened 2019-10-25
Run lorri daemon as a service on Linux and macOS

### LINKED ISSUE #96: lorri daemon: systemd service
I've started using the branch with the [`lorri daemon` PR](https://github.com/target/lorri/pull/81) and it's pretty cool! 1 annoyance it carries over from regular lorri is that it still requires an extra sacrificial terminal... with a shell.nix (any shell.nix it seems?) to launch the daemon and remain open. At least it's only 1 as opposed to many.

Anyway, it'd be nice if it could run as a (systemd, etc.) service.

Thanks, everyone! I've been really enjoying lorri

curiousleo:
**[`lorri`, including a user service definition, is now in `nixpkgs`](https://github.com/NixOS/nixpkgs/pull/72889) :tada:**

---
Addresses some of #96. `lorri.socket`, `lorri.service` and `lorri.nix` are mostly taken from comments on that issue.

[`contrib/README.md` preview](https://github.com/curiousleo/lorri/blob/contrib/contrib/README.md).

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,0 +1,102 @@
+# How to start the lorri daemon as a service
+
+This guide shows you how to run `lorri daemon` as a service. The exact steps
+depend on your operating system and general setup. Currently, we have
+instructions for these setups:
+
+- [Run `lorri daemon` on NixOS](#run-lorri-daemon-on-nixos)
+- [Run `lorri daemon` on Linux with
+  home-manager](#run-lorri-daemon-on-linux-with-home-manager)
+- [Run `lorri daemon` on Linux with just
+  systemd](#run-lorri-daemon-on-linux-with-just-systemd)
+- [Run `lorri daemon` on macOS with
+  Nix](#run-lorri-daemon-on-macOS-with-nix)
+
+## Run `lorri daemon` on NixOS
+
+Add [`services.lorri.enable =
+true;`](https://nixos.org/nixos/options.html#services.lorri.enable) to your
+`configuration.nix`, then rebuild your system:
+
+```console
+$ sudo nixos-rebuild switch
+```
+
+The lorri daemon will now be started on demand by systemd. See [Verify the
+setup](#verify-the-setup) to check that everything works as expected.
+
+## Run `lorri daemon` on Linux with home-manager
+
+Add [`services.lorri.enable =
+true;`](https://rycee.gitlab.io/home-manager/options.html#opt-services.lorri.enable)
+to your `home.nix`, then switch to the new configuration:
+
+```console
+$ home-manager switch
+```
+
+The lorri daemon will now be started on demand by systemd. See [Verify the
+setup](#verify-the-setup) to check that everything works as expected.
+
+## Run `lorri daemon` on Linux with just systemd
+
+Here we'll set up a [systemd] socket and service file manually.
+
+<details>
+<summary>What's the purpose of the systemd socket? How does systemd know when
+to start the daemon "on demand"?</summary>
+<p>lorri clients, like the `direnv` integration, talk to the daemon via a Unix
+socket at a well-known location. [`lorri.socket`] tells systemd to start the
+systemd service defined in [`lorri.service`] the first time a client attempts
+to connect to this socket.</p>
+</details>
+
+If your `lorri` binary is not in `~/.nix-profile/bin/lorri`, please change the
+`ExecStart=` setting in `lorri.service` to the correct location.
+
+Install [`lorri.socket`] and [`lorri.service`] and make systemd listen on the
+daemon socket:
+
+```console
+$ mkdir -p ~/.config/systemd/user && \
+    cp contrib/lorri.{socket,service} ~/.config/systemd/user/ && \
+    systemctl --user daemon-reload && \
+    systemctl --user enable --now lorri.socket
+```
+
+The lorri daemon will now be started on demand by systemd. See [Verify the
+setup](#verify-the-setup) to check that everything works as expected.
+
+## Run `lorri daemon` on macOS with Nix
+
+Please take a look at [@pawlowsklalaex]'s [suggested
+`darwin-configuration.nix`](https://github.com/target/lorri/issues/96#issuecomment-545152525).
+
+## Verify the setup
+
+In this section, we'll see how to check that the `lorri daemon` setup actually
+works as intended.
+
+### systemd
+
+On a systemd-based system, you should get the following:
+
+```console
+$ systemctl --user is-enabled lorri.socket
+enabled
+$ systemctl --user is-active lorri.socket
+active
+```
+
+### launchd
+
+On macOS, use this command to check the status of the lorri daemon:
+
+```console
+$ launchctl list | grep lorri
+```
+
+[systemd]: https://www.freedesktop.org/wiki/Software/systemd/
+[`lorri.socket`]: ./lorri.socket
+[`lorri.service`]: ./lorri.service
+[@pawlowsklalaex]: https://github.com/pawlowskialex
--- a/contrib/lorri.service
+++ b/contrib/lorri.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Lorri Daemon
+Requires=lorri.socket
+After=lorri.socket
+
+[Service]
+ExecStart=%h/.nix-profile/bin/lorri daemon
+PrivateTmp=true
+ProtectSystem=strict
+ProtectHome=read-only
+Restart=on-failure
--- a/contrib/lorri.socket
+++ b/contrib/lorri.socket
@@ -1,0 +1,9 @@
+[Unit]
+Description=Socket for Lorri Daemon
+
+[Socket]
+ListenStream=%t/lorri/daemon.socket
+RuntimeDirectory=lorri
+
+[Install]
+WantedBy=sockets.target


curiousleo:
Works For Me (TM) on NixOS 19.09.

contrib/README.md
```diff
@@ -0,0 +1,70 @@
...
+$ systemctl --user daemon-reload
+$ systemctl --user list-unit-files | grep lorri
+lorri.service                         static  
+lorri.socket                          disabled
+```
+
+Now start the socket:
+
+```bash
+$ systemctl --user start lorri.socket
```
curiousleo: Use `systemctl --user enable --now lorri.socket` instead?

curiousleo:
I'm looking into how we can check these files in CI. There's `systemd-analyze verify` for the `.socket` and `.service` files: http://manpages.ubuntu.com/manpages/bionic/man1/systemd-analyze.1.html#examples%20for%20verify

contrib/README.md
curiousleo: No: this socket is not mean to be `enable`d. We can start it here; on system reboot it will be started again because it is `WantedBy=sockets.target`.

curiousleo:
> I'm looking into how we can check these files in CI. There's `systemd-analyze verify` for the `.socket` and `.service` files: http://manpages.ubuntu.com/manpages/bionic/man1/systemd-analyze.1.html#examples%20for%20verify

Things I've learned:
- Currently we don't set the `dist` in `.travis.yml`. [The default is `xenial`](https://docs.travis-ci.com/user/reference/overview/) (Ubuntu 16.04).
- Even on `bionic` (Ubuntu 18.04, the latest `dist` available on Travis CI), [`systemd` is only at version 237.](https://packages.ubuntu.com/bionic/systemd)
- Somehow, `systemd-analyze` still [reports its version as 229](https://travis-ci.org/target/lorri/jobs/603926080#L324) on `bionic` (I set `dist: bionic` in that CI run).

Since the names of systemd unit settings have changed, running `systemd-analyze verify` in CI is difficult: `lorri.socket` and `lorri.service` do actually work fine on systemd 243 (the version shipped in NixOS 19.09), but older versions of `systemd-analyze verify` report errors.

curiousleo:
PTAL

flokli:
I'm a big fan of adding these files into `contrib/`, and refer those in a NixOS module for example.

I don't really like we need to document multiple ways on how to get it set up - can't we please just release Lorri, get it into nixpkgs, and provide a NixOS module using the unit files in `contrib/`? :laughing:

ocharles:
I think the NixOS module should also install the `lorri` binary so it's available to all users, or `eval "$(lorri direnv)"` won't work in `.envrc` files (as lorri can't be found)

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
...
+      description = "Socket for Lorri Daemon";
+      enable = true;
+      wantedBy = [ "sockets.target" ];
+      socketConfig = {
+        ListenStream = "%t/${socketPath}";
+        RuntimeDirectory = "lorri";
+      };
+    };
+
+    systemd.user.services.lorri = {
```
ocharles: The `PATH` that I end up with for this doesn't have access to `nix-instantiate`, so `lorri` doesn't actually work.

curiousleo:
Thank you @ocharles for trying this out! I believe that in order to address the issues you've raised, we'll have to think about the Nix packaging of `lorri` more generally - that is, actually have a packaged `.nix` file that makes sure `nix` is available as a runtime dependency, and then have the `.nix` file describing the service integration depend on that. (They could be the same file.)

Note to self: check out https://eipi.xyz/blog/packaging-a-rust-project-for-nix/

curiousleo:
Quick update - I'm working on getting `lorri` included in `nixpkgs` right now: https://github.com/NixOS/nixpkgs/pull/72889, including the service definition. That will eliminate the manual steps on NixOS at least, and it will make it easier to include the service definition in `home-manager` (i.e. https://github.com/rycee/home-manager/pull/799 could be re-opened).

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
+{ config, systemd, ... }:
+
+let
+  lorri = import
+    (fetchTarball "https://github.com/target/lorri/tarball/rolling-release")
+    { };
+  socketUnit = "lorri";
+  socketPath = "lorri/daemon.socket";
```
grahamc: Let's not have these as variables, and instead just use "lorri" and "lorri/daemon.socket". They're not really variable, just trying to reduce a bit of duplication I think in a way that doesn't get us much.

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
...
+      socketConfig = {
+        ListenStream = "%t/${socketPath}";
+        RuntimeDirectory = "lorri";
+      };
+    };
+
+    systemd.user.services.lorri = {
+      description = "Lorri Daemon";
+      requires = [ "${socketUnit}.socket" ];
+      after = [ "${socketUnit}.socket" ];
```
grahamc:
```suggestion
      after = [ "${socketUnit}.socket" ];
      path = with pkgs; [ nix ];
```
 
probably need some others too.

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
+{ config, systemd, ... }:
```
grahamc:
```suggestion
{ config, pkgs, ... }:
```

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
...
+
+let
+  lorri = import
+    (fetchTarball "https://github.com/target/lorri/tarball/rolling-release")
+    { };
+  socketUnit = "lorri";
+  socketPath = "lorri/daemon.socket";
+in {
+  config = {
+    systemd.user.sockets.${socketUnit} = {
```
grahamc:
```suggestion
    systemd.user.sockets.lorri = {
```

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
...
+      description = "Socket for Lorri Daemon";
+      enable = true;
+      wantedBy = [ "sockets.target" ];
+      socketConfig = {
+        ListenStream = "%t/${socketPath}";
```
grahamc:
```suggestion
        ListenStream = "%t/lorri/daemon.socket";
```

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
...
+        RuntimeDirectory = "lorri";
+      };
+    };
+
+    systemd.user.services.lorri = {
+      description = "Lorri Daemon";
+      requires = [ "${socketUnit}.socket" ];
```
grahamc:
```suggestion
      requires = [ "lorri.socket" ];
```

contrib/lorri.nix
```diff
@@ -0,0 +1,38 @@
+{ config, systemd, ... }:
+
+let
+  lorri = import
+    (fetchTarball "https://github.com/target/lorri/tarball/rolling-release")
+    { };
+  socketUnit = "lorri";
```
grahamc:
```suggestion
```

contrib/lorri.nix
grahamc:
```suggestion
```

curiousleo: That's what flokli said :) https://github.com/NixOS/nixpkgs/pull/72889#discussion_r345109924

curiousleo:
Thanks for the review @grahamc! Note that now `lorri` has landed in nixpkgs (https://github.com/NixOS/nixpkgs/pull/72889) :tada: I think this PR can be simplified quite a lot. I'm not exactly sure what the instructions should be now, to be honest, given that `lorri` is now in `nixpkgs/master`, but not yet in `unstable` nor in any `stable` branch.

contrib/README.md
```diff
@@ -0,0 +1,121 @@
...
+  home-manager](#run-lorri-daemon-on-linux-with-home-manager)
+- [Run `lorri daemon` on Linux with just
+  systemd](#run-lorri-daemon-on-linux-with-just-systemd)
+- [Run `lorri daemon` on macOS with
+  home-manager](#run-lorri-daemon-on-macOS-with-home-manager)
+
+## Run `lorri daemon` on NixOS
+
+On NixOS, first copy [`lorri.nix`] into your NixOS configuration
+directory:
```
Profpatsch: We have the NixOS service in nixpkgs now, so we can get rid of this methinks?

curiousleo: I partially agree. This should all become "put `services.lorri.enable = true;` in your `configuration.nix`. Unfortunately right now that only works for those brave enough to run NixOS `master` (does anyone?) - so that will only be widely applicable once `lorri` is in NixOS stable, methinks.

In the meantime, I'm not sure what we should suggest. Perhaps we can point to the package and service definition on nixpkgs and ask people to get them from there? Either way, I agree with you that the package and service nix files don't belong in this PR anymore now that improved versions of them are in nixpkgs.

flokli: I think it's fine to backport non-breaking changes (additions) to the stable release branch, too.

For example, `iwd` support for `NetworkManager` was backported to 19.09.

Profpatsch: Okay, I think we only improve the situation if we backport (right now the `lorri` attribute in nixpkgs 19.09 throws an error).

Profpatsch: https://github.com/NixOS/nixpkgs/pull/73451

flokli:
With NixOS/nixpkgs#73451 merged aswell, we should direct users to just set `services.lorri.enable = true;`.

curiousleo:
> I pinged at [rycee/home-manager#799](https://github.com/rycee/home-manager/pull/799), it would be nice if we could have a home-manager module aswell, which we could refer to.

`services.lorri.enable` has landed in `home-manager`: https://github.com/rycee/home-manager/blob/286dd9b3088298e5a4625b517f8e72b1c62e4f74/modules/services/lorri.nix - I'll adjust the `home-manager` instructions in the README accordingly.

curiousleo:
PTAL - I'd like to merge this now (it's better to have this than not to have this) and then revisit it as part of https://github.com/target/lorri/issues/231.

curiousleo:
> Let’s merge after squashing. Since you’re gonna re-work parts anyway, I don’t see why we should block the merge.

:+1: I've squashed the commits.

PR #193, opened 2019-10-27
default.nix: use naersk, remove carnix

Profpatsch:


grahamc:
Not right now, no. I want to stick with carnix for now.

curiousleo:
Relevant for lorri's use case: https://github.com/nmattia/naersk/pull/46

PR #194, opened 2019-10-28
ping_: canonicalize path

### LINKED ISSUE #192: `lorri ping_` should canonicalise relative paths
`lorri ping_` should canonicalise paths before passing them on to the daemon. The current workaround is `lorri ping_ $(readlink -f shell.nix)`, but we could just as well do the canonicalisation in `lorri ping_`.

curiousleo:
Fixes #192.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,7 +1,6 @@
 //! Defines the CLI interface using structopt.
 
 use std::path::PathBuf;
-use NixFile;
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "lorri")]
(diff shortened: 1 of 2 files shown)


curiousleo:
Thanks for the review @Profpatsch!

> I concur. Though let’s just `impl` canonicalize for `NixFile` itself (so we have a good listing of what use-cases there are for `NixFile` close to its definition.

I only just noticed that there was a pattern for other commands where the struct for parsing the arguments holds a `PathBuf` and the validation happens in `main.rs`. I've just adopted this approach for `Ping_` for the time being - that fixes the bug without adding complexity.

> I also thought about making `NixFile` absolute by construction, that could be another approach, though I’d have to experiment with that.

I played around with that for a bit. It looks to me like a few `TryFrom` declarations (from `&OsStr`, `PathBuf` etc.) could be a good way to codify the possibly-failing creation of `NixFile`.

curiousleo:
PTAL

Profpatsch:
Looks good! LGTM.

PR #198, opened 2019-11-04
lib/NixFile make absolute by construction

Profpatsch:
This expands on https://github.com/target/lorri/pull/194 a bit by making `NixFile` always be an absolute path, correct by construction.

It then factors out the absolute path part to also add it to `cas.rs` and `constants/Paths`.
We should propagate further into `project.rs` and `roots.rs`, but I’ve left that as a TODO for now.

Most of these changes are rather mechanical (and mostly trivial).

cc @curiousleo

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -4,11 +4,14 @@ extern crate directories;
 
 use self::directories::ProjectDirs;
 use crate::cas::ContentAddressable;
+use crate::AbsPathBuf;
 use std::path::{Path, PathBuf};
 
 /// Path constants like the GC root directory.
 pub struct Paths {
+    // TODO: make AbsPathBuf
     gc_root_dir: PathBuf,
+    // TODO: make AbsPathBuf
     daemon_socket_file: PathBuf,
     cas_store: ContentAddressable,
 }
@@ -32,7 +35,7 @@ pub enum PathsInitError {
     /// The CAS creation failed.
     #[allow(missing_docs)]
     CasCantBeCreated {
-        cas_dir: PathBuf,
+        cas_dir: AbsPathBuf,
         err: std::io::Error,
     },
 }
@@ -51,7 +54,15 @@ impl Paths {
             // fall back to the cache dir on non-linux
             .unwrap_or_else(|| pd.cache_dir())
             .to_owned();
-        let cas_dir = pd.cache_dir().join("cas");
+
+        // TODO: return as good error value
+        assert!(
+            pd.cache_dir().is_absolute(),
+            "Your cache directory is not an absolute path! It is: {}",
+            pd.cache_dir().display()
+        );
+        let abs_cache_dir = AbsPathBuf::new_unchecked(pd.cache_dir().to_owned());
+        let cas_dir = abs_cache_dir.join("cas");
         Ok(Paths {
             gc_root_dir: create_dir(gc_root_dir.clone()).map_err(|err| {
                 PathsInitError::GcRootsDirectoryCantBeCreated { gc_root_dir, err }
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -46,13 +46,14 @@ impl Paths {
             std::fs::create_dir_all(&dir).and(Ok(dir))
         };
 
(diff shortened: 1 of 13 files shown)


src/constants.rs
```diff
@@ -18,33 +18,49 @@ impl Paths {
...
             .expect("Could not determine lorri project/cache directories, please set $HOME");
-        let create_dir = |dir: PathBuf| -> std::io::Result<PathBuf> {
+        let create_dir = |dir: AbsPathBuf| -> std::io::Result<AbsPathBuf> {
             std::fs::create_dir_all(&dir).and(Ok(dir))
         };
+
+        // TODO: return as good error value
+        assert!(
+            pd.cache_dir().is_absolute(),
+            "Your cache directory is not an absolute path! It is: {}",
```
curiousleo: Is there anything the user could do about this error message when they see it? `pd` comes from `ProjectDirs::from`, so it seems to me like that method should ensure this is a "good" path. Perhaps `ProjectDirs::cache_dir` could return an `AbsPathBuf`?

src/constants.rs
```diff
@@ -18,33 +18,49 @@ impl Paths {
...
+
+        let runtime_dir = pd
+            .runtime_dir()
+            // fall back to the cache dir on non-linux
+            .unwrap_or_else(|| pd.cache_dir())
+            .to_owned();
+        // TODO: return as good error value
+        assert!(
+            runtime_dir.is_absolute(),
+            "Your runtime directory is not an absolute path! It is: {}",
```
curiousleo: Same as above

src/main.rs
```diff
@@ -44,48 +44,56 @@ fn main() {
...
-fn get_shell_nix(shellfile: &PathBuf) -> Result<NixFile, ExitError> {
+/// Reads a nix filename given by the user and either returns
+/// the `NixFile` type or exists with a helpful error message
+/// that instructs the user how to write a minimal `shell.nix`.
+fn find_nix_file(shellfile: &PathBuf) -> Result<NixFile, ExitError> {
     // use shell.nix from cwd
-    Ok(NixFile::from(locate_file::in_cwd(&shellfile).map_err(
-        |_| {
+    Ok(NixFile::from(
+        // `::in_cwd` is guaranteed to return an absolute path
```
curiousleo: Shouldn't `in_cwd` just return an `AbsPathBuf` in that case?

Profpatsch: Good idea, changed in last commit.

src/constants.rs
Profpatsch: `ProjectDirs::cache_dir` comes from a different library. I opted for a `panic!()` for now, because if the paths return by `ProjectDirs` are not absolute, we have a problem (because we want to ensure that nix files are absolute).

Profpatsch:
> I don't understand this PR

We must ensure that `.nix` files are always absolute. Otherwise the daemon cannot function correctly. While adding that guarantee I noticed that we should guarantee it for the `cas` module as well, because we generate nix paths from that, which are read by the daemon. So I factored the shared preconditions into `AbsPathBuf`.
This follows the approach of [parse, don’t validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) (actually this commit is from before that blogpost, but that explains the ideas quite well. An older and more concise description is in [Matt Parson’s blog](https://www.parsonsmatt.org/2017/10/11/type_safety_back_and_forth.html).

> the only way to construct one is via `new_unchecked`. This means there is no actual verification "by construction" that the contained PathBuf is actually absolute.

I added a `new` function which does a check, and added it at the borders where we cannot be entirely sure that a path is absolute.

Profpatsch:
rebased on master.

gilligan:
@grahamc you had blocked this a good while ago. Does your opinion still hold? Should this be abandoned or is there interest on either side to pick this up again?

Profpatsch:
I’m sure it’s going to prevent bugs in the future, but my mental health does not depend on it being merged :)

Profpatsch:
Merged in https://github.com/nix-community/lorri/pull/14

PR #202, opened 2019-11-06
constants/Paths: Split IO errors into enum for easier debugging

Profpatsch:
what it says on the tin. Good error messages if something in the user environment is not right.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -13,24 +13,57 @@ pub struct Paths {
     cas_store: ContentAddressable,
 }
 
+/// Everything that can happen when creating `Paths`.
+/// Mostly filesystem access problems.
+#[derive(Debug)]
+pub enum PathsInitError {
+    /// The `gc_root_dir` creation failed.
+    #[allow(missing_docs)]
+    GcRootsDirectoryCantBeCreated {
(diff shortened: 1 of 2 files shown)


src/constants.rs
```diff
@@ -13,24 +13,57 @@ pub struct Paths {
     cas_store: ContentAddressable,
 }
 
+/// Everything that can happen when creating `Paths`.
+/// Mostly filesystem access problems.
+#[derive(Debug)]
+pub enum PathsInitError {
+    /// The `gc_root_dir` creation failed.
+    #[allow(missing_docs)]
+    GcRootsDirectoryCantBeCreated {
```
curiousleo: I suggest
- `InitPathsError::CreateGcRootsDir`
- `InitPathsError::CreateSocketDir`
- `InitPathsError::CreateCas`

I think this fits better with how errors are named in the [standard library](https://doc.rust-lang.org/std/all.html?search=error).

Profpatsch: I’d like for them to be as descriptive as possible, so the user knows what the problem was. `CreateGcRootsDir` doesn’t really give the same air as `GcRootsDirectoryCantBeCreated`. (They are not really intended to be consumed programmatically anyway.)

PR #207, opened 2019-11-09
shell.nix: minimize shell when not `isDevelopmentShell`

Profpatsch:
Remove development specific environment variables from derivation.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,8 +1,13 @@
-{ pkgs ? import ./nix/nixpkgs.nix { enableMozillaOverlay = true; }
-, isDevelopmentShell ? true }:
+{
+  # Pull in tools & environment variables that are only
+  # required for interactive development (i.e. not necessary
+  # on CI). Only when this is enabled, Rust nighly is used.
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -1,8 +1,9 @@
-{ pkgs ? import ./nix/nixpkgs.nix { enableMozillaOverlay = true; }
-, isDevelopmentShell ? true }:
+{ isDevelopmentShell ? true
```
curiousleo: Perhaps you could add a comment that explains the meaning or effect of setting / unsetting `isDevelopmentShell`.

grahamc:
why?

shell.nix
Profpatsch: Done, thanks.

Profpatsch:
> why?

Unnecessary to pull in rust unstable on CI.

Profpatsch:
I’ll just merge, we can revert should it make any problems, but the CI does have to set up less stuff if the shell is minimal.

PR #208, opened 2019-11-10
Codify install instructions

chrissound:
Do note, this does hard code the branch to master. But I think it would be fine.


curiousleo:
LGTM

Profpatsch:
Thanks, improved it slightly.

PR #209, opened 2019-11-10
direnv directly from nixos rather than github

chrissound:
I'm not 100% sure of this change - but I think it's more ideal as direnv is on direnv-2.20.1 on nixos 19.09 currently (I'm guessing it was an older version on the nixos channel previously).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,7 @@ For Nix on Linux or macOS, you can install the needed version of
 direnv with:
 
 ```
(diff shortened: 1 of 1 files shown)


curiousleo:
LGTM

Profpatsch:
The linked nix expression in our repo does a fallback (it uses direnv from nixpkgs if its newer than what we require, the minimal working version otherwise).

PR #210, opened 2019-11-10
Readme minor change - Create file via echo command

chrissound:


curiousleo:
Thanks @chrissound! My suggestion would be to shorten the text a bit, i.e. perhaps _replace_

---
Create a file named `.envrc` in your project's root with the contents:

```
eval "$(lorri direnv)"
```
---

by:

---
Create a file named `.envrc` in your project's root as follows:
```
echo 'eval "$(lorri direnv)"' > .envrc
```
---

Profpatsch:
Thanks, but I don’t think we can assume that people are in their project’s root, and they will have a bad time if they just copy & paste some command and are not in the right directory.

PR #214, opened 2019-11-12
Ping should always trigger build

### LINKED ISSUE #212: `lorri ping_` doesn’t trigger reevaluation if the daemon watches the nix file already
I noticed that if I run `lorri direnv` (via the direnv integration), the daemon sometimes does not start a build.

More specifically, if I remove the `shell_gc_root` (which the daemon prints in its `Started` messages), the environment won’t be rebuild if I run `lorri ping_ /path/to/shell.nix`, as long as the daemon is still running. If I restart the daemon, it is rebuilt after `ping_`.

The ping is received by the daemon here:

https://github.com/target/lorri/blob/a0ff0a51f742f1e73f4adefd0d46c67c6032016c/src/ops/daemon.rs#L42-L44
it is written to `accept_messages_tx` and read from the reading end further down:

https://github.com/target/lorri/blob/a0ff0a51f742f1e73f4adefd0d46c67c6032016c/src/ops/daemon.rs#L60-L74

The problem lies in `daemon.add()`, which has a different semantics than what I thought when writing this code:

https://github.com/target/lorri/blob/a0ff0a51f742f1e73f4adefd0d46c67c6032016c/src/daemon.rs#L64-L80

It adds projects to the daemon, but only if the project was not added yet (`or_insert_with`).

Instead, we need to be able to influence the `build_loop`s somehow, which are running inside threads. I propose to add a channel going from the daemon to each `build_loop`s and extending the hash map by the `tx` going to each builder:

https://github.com/target/lorri/blob/a0ff0a51f742f1e73f4adefd0d46c67c6032016c/src/daemon.rs#L29-L31

Then we can forward a message to the build_loop that a build should be queried, even if no file has changed in the meantime.

curiousleo:
Fixes #212.

Pretty much follows [@Profpatsch's suggested approach](https://github.com/target/lorri/issues/212#issue-521493053). I've tested this by following the [repro steps](https://github.com/target/lorri/issues/212#issuecomment-552917458).

This is a "minimal diff" fix: it fixes the issue whilst touching the minimum amount of code. I believe though that the build loop is now too complex, so I think this should be followed up by a refactor to make the code more readable and maintainable.

**Closed in favour of https://github.com/target/lorri/pull/221**

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -25,10 +25,15 @@ pub struct IndicateActivity {
     pub nix_file: NixFile,
 }
 
+struct Handler {
+    tx: mpsc::Sender<()>,
+    _handle: std::thread::JoinHandle<()>,
+}
+
 /// Keeps all state of the running `lorri daemon` service, watches nix files and runs builds.
 pub struct Daemon {
     /// A thread for each `BuildLoop`, keyed by the nix files listened on.
-    handler_threads: HashMap<NixFile, std::thread::JoinHandle<()>>,
+    handler_threads: HashMap<NixFile, Handler>,
     /// Sending end that we pass to every `BuildLoop` the daemon controls.
     // TODO: this needs to transmit information to identify the builder with
     build_events_tx: mpsc::Sender<crate::build_loop::Event>,
@@ -64,19 +69,25 @@ impl Daemon {
     /// Add nix file to the set of files this daemon watches
     /// & build if they change.
     pub fn add(&mut self, project: Project) {
-        let tx = self.build_events_tx.clone();
+        let (tx, rx) = mpsc::channel();
+        let build_events_tx = self.build_events_tx.clone();
 
         self.handler_threads
             .entry(project.nix_file.clone())
-            .or_insert_with(|| {
-                std::thread::spawn(move || {
+            .or_insert_with(|| Handler {
+                tx,
+                _handle: std::thread::spawn(move || {
                     let mut build_loop = BuildLoop::new(&project);
 
                     // cloning the tx means the daemon’s rx gets all
                     // messages from all builders.
-                    build_loop.forever(tx);
-                })
-            });
+                    build_loop.forever(build_events_tx, rx);
+                }),
+            })
+            // Notify the handler, whether or not it was newly added
+            .tx
+            .send(())
+            .unwrap();
     }
 }
 
--- a/src/ops/watch.rs
+++ b/src/ops/watch.rs
@@ -38,7 +38,7 @@ fn main_run_forever(project: Project) -> OpResult {
     let build_thread = {
         thread::spawn(move || {
             let mut build_loop = BuildLoop::new(&project);
-            build_loop.forever(tx);
+            build_loop.forever(tx, channel().1);
         })
     };
 
--- a/src/build_loop.rs
+++ b/src/build_loop.rs
@@ -10,7 +10,7 @@ use crate::project::roots::Roots;
 use crate::project::Project;
 use crate::watch::{DebugMessage, RawEventError, Reason, Watch};
 use std::path::PathBuf;
-use std::sync::mpsc::{channel, Sender};
+use std::sync::mpsc::{channel, Receiver, Sender};
 
 /// Builder events sent back over `BuildLoop.tx`.
 #[derive(Clone, Debug)]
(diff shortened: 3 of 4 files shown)


src/daemon.rs
```diff
@@ -64,19 +69,25 @@ impl Daemon {
...
-                    build_loop.forever(tx);
-                })
-            });
+                    build_loop.forever(build_events_tx, rx);
+                }),
+            })
+            // Notify the handler, whether or not it was newly added
+            .tx
+            .send(())
+            .unwrap();
```
curiousleo: This is the fix, basically.

src/ops/watch.rs
```diff
@@ -38,7 +38,7 @@ fn main_run_forever(project: Project) -> OpResult {
     let build_thread = {
         thread::spawn(move || {
             let mut build_loop = BuildLoop::new(&project);
-            build_loop.forever(tx);
+            build_loop.forever(tx, channel().1);
```
curiousleo: I'm assuming here that `ping_` is not expected to work with `watch`.

grahamc:
We shouldn't issue an evaluation/build on every ping. we will receive a ping every time the user changes buffers in emacs, for example. Instead, let's check  to see if the GC root exists. If not, start the build.

curiousleo:
> We shouldn't issue an evaluation/build on every ping. we will receive a ping every time the user changes buffers in emacs, for example. Instead, let's check to see if the GC root exists. If not, start the build.

In this context, it sounds sensible to me to abstain from rebuilding on every ping. I understand what you're saying as:
```
on ping:
    if !gc_root.exists()
        rebuild()
```

Is that correct?

This contradicts https://github.com/target/lorri/issues/212#issuecomment-553667616 a little bit, which I understand as "watch the GC root using notify". In that case the build loop doesn't need to care about pings because it will trigger rebuilds immediately anyway when the GC root is removed.

grahamc:
Right. Not sure why, but I replied to your reply over email and GitHub seems to not have posted it. Essentially no need to watch the GC roots because we are the only ones who should be touching them. IN the case one is deleted, it should be fine to realize that as soon as the next ping is received (which is often if they're in an editor.)


```
on ping:
    if !gc_root.exists()
        rebuild()
```

is exactly right as far as I'm thinking

Profpatsch:
We have to distinguish between `lorri ping_` and “direnv was run by switching buffer in Emacs” me thinks. The first ono is intended to always trigger an evaluation, the second one just incidentally uses the first one at the moment, but we should probably change that.

curiousleo:
> We have to distinguish between `lorri ping_` and “direnv was run by switching buffer in Emacs” me thinks. The first ono is intended to always trigger an evaluation, the second one just incidentally uses the first one at the moment, but we should probably change that.

I'm not 100% sure I understand this, but the relevant question for me is: would

```
on ping:
    if !gc_root.exists():
        rebuild()
```

work for all the cases in which pings are sent, or would it break stuff?

curiousleo:
> We have to distinguish between `lorri ping_` and “direnv was run by switching buffer in Emacs” me thinks. The first ono is intended to always trigger an evaluation, the second one just incidentally uses the first one at the moment, but we should probably change that.

Ah, are you saying that `lorri ping_` should always _force_ a rebuild?

Profpatsch:
> Ah, are you saying that `lorri ping_` should always _force_ a rebuild

It should always queue an evaluation (& rebuild if necessary)



> whereas a ping from direnv should only lead to a build if there is no GC root?

Good question, I haven’t thought about that, but from what Graham wrote above I infered that’s the preferred modus operandi for him.

curiousleo:
> It should always queue an evaluation (& rebuild if necessary)

I just want to understand the use case here. If we're already watching all the build inputs using `notify`, under what circumstances would this be useful? My mental model is that a `ping_`-triggered rebuild will always be a no-op since any relevant changes will already have been picked up because they'll have triggered a rebuild via `notify`. But I may well be missing something here.

grahamc:
`ping` should not force a rebuild, it should do the exact same thing `lorri direnv` does, minus the direnv goop. If there is reason for forcing an evaluation or build, let's open up a new issue describing the motivation and desired behavior. The reason this PR is looking like a reasonable step forward is: indeed, if the gc root is no longer present and a ping is received ... we should put it back.

curiousleo:
Closing in favour of https://github.com/target/lorri/pull/221.

PR #217, opened 2019-11-13
Upgrade notify, switch to crossbeam_channel

curiousleo:
**Prerequisite for [a PR to simplify the build loop.](https://github.com/curiousleo/lorri/pull/3)** This PR contains zero logic changes, it's purely a search-and-replace affair. I've split it out because I think it should be quick and easy to review.

What's in this PR?
- Upgrade `notify` crate from v4 to v5
- Switch from `mpsc` channels to `crossbeam_channel` (used by the new version of `notify`)

**Why upgrade `notify`?**
[Notify v4 has been abandoned by its author.](https://docs.rs/crate/notify/4.0.14) Notify v5 is what the new maintainers are working on now. It's still in pre-release, but AFAICT that just means the API hasn't been 100% stabilised yet. So we pin the exact version to avoid surprises.

**Why switch to `crossbeam_channel`?**
`crossbeam_channel` is a more performant drop-in replacement for `mpsc` with more features.
- `notify` v5 depends on it
- `select!` [has recently been removed from `mpsc`](https://github.com/rust-lang/rust/pull/60921) (issue [here](https://github.com/rust-lang/rust/issues/27800)). The ability to `select`, i.e. to wait on multiple channels at once, is IMO a crucial part of any channel API. [`crossbeam_channel` has `select!`.](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/macro.select.html) It seems that the ecosystem as a whole is moving from `mpsc` to `crossbeam_channel`, see for example [`async-std`](https://github.com/async-rs/async-std/blob/f611ceccc87aaf6b6cfb5bdee8bfffaffee5498f/Cargo.toml#L57).


Cargo.toml
```diff
@@ -9,24 +9,25 @@ license = "Apache-2.0"
 edition = "2018"
 
 [dependencies]
```
curiousleo: I couldn't resist sorting the dependencies alphabetically.

curiousleo: The only changes here are:
- changed `notify = "4.0.6"` to `notify = "5.0.0-pre.1"
- added `crossbeam-channel = "0.3.8"`

curiousleo:
PTAL

Profpatsch:
> It's still in pre-release, but AFAICT that just means the API hasn't been 100% stabilised yet. So we pin the exact version to avoid surprises.

Not entirely sure we want to update to a 5.0 pre-release yet, especially since v4 is [still getting bug fix updates](https://github.com/notify-rs/notify/releases).


> The ability to `select`, i.e. to wait on multiple channels at once, is IMO a crucial part of any channel API.

Agreed.

It sounds like you need the ability to `select` on channels for the `build_loop` update? If so, then let’s do it.

curiousleo:
> It sounds like you need the ability to `select` on channels for the `build_loop` update? If so, then LGTM.

Yeah, the idea is for the build loop to become essentially
```rust
loop {
    select! {
        // receive simultaneously from various channels that might trigger a rebuild
    }
}
```

Profpatsch:
> Yeah, the idea is for the build loop to become essentially
> 
> ```rust
> loop {
>     select! {
>         // receive simultaneously from various channels that might trigger a rebuild
>     }
> }
> ```

Good thinking, that’s my biggest issue with https://github.com/target/lorri/pull/214 as it stands, that `build_loop` is becoming too complicated.

PR #221, opened 2019-11-14
Fix #212, simplify build loop

### LINKED ISSUE #212: `lorri ping_` doesn’t trigger reevaluation if the daemon watches the nix file already

curiousleo:
This is a better fix for #212 than my previous "minimal change" fix in https://github.com/target/lorri/pull/214. It builds on top of https://github.com/target/lorri/pull/217.

**What's in this PR?**
- fixes #212 based on this idea: https://github.com/target/lorri/pull/214#issuecomment-553963285
- simplifies the build loop in `BuildLoop::forever`
- replaces our own debouncing logic by `notify`'s (but keeping all the tests to ensure we get the same behaviour)

**Changes to `BuildLoop::forever`**
This PR changes both what the build loop does and how it does it.

The behavioural change is from
```
on notify {
    if notification_relevant() {
        rebuild()
    }
}
```
to
```
on notify {
    if notification_relevant() {
        rebuild()
    }
}
on ping {
    if !gc_root.exists() {
        rebuild()
    }
}
```
That is, a rebuild is now _also_ triggered if no GC root exists when a ping is received. This fixes #212.

The implementation also changes: Previously the main loop blocked on `self.watch.wait_for_change()`. Now it uses `chan::select!` to wait on a message, which comes either from a ping or from a notification.

**Using `notify`'s built-in debouncing logic**
Switching from [`Watcher::new_immediate`](https://docs.rs/notify/5.0.0-pre.1/notify/trait.Watcher.html#tymethod.new_immediate) to [`Watcher::new`](https://docs.rs/notify/5.0.0-pre.1/notify/trait.Watcher.html#tymethod.new) lets us use `notify`'s own debouncing logic. In combination with the switch from timout-based, blocking calls inside the build loop to a message-based approach, this allows for a substantial simplification of `Watch` as well as the removal of `FilterTimeoutIterator`.


curiousleo:
PTAL

src/build_loop.rs
```diff
@@ -63,43 +63,65 @@ impl<'a> BuildLoop<'a> {
...
 
-        send(Event::Started(Reason::ProjectAdded(
+        // The project has just been added, so run the builder in the first iteration
+        let mut reason = Some(Event::Started(Reason::ProjectAdded(
             self.project.nix_file.clone(),
         )));
+        let mut output_paths = None;
+
+        // Drain pings initially: we're going to trigger a first build anyway
+        rx_ping.try_iter().for_each(drop);
```
grahamc: Is it possible this will drain pings for other projects, too?

curiousleo: No: `rx_ping` is created by `Daemon::add`, so there is one channel per project. For each project, there is a `Handler` in `Daemon::handler_threads` which contains the sender `tx` corresponding to the `rx_ping` receiver. It's somewhat implicit, but I hope that this structure makes it clear that it is _intended_ to have one of those channels per project.

curiousleo:
PTAL

PR #232, opened 2019-11-19
[WIP] readDir: do not watch directories recursively

### LINKED ISSUE #152: Improve readDir watches
Lorri seems to watch directories recursively when it detects `builtins.readDir`. This seems unnecessary because `readDir` only returns the direct children. Any use of those children should be detectable with existing means.

curiousleo:
**Do not merge.** See https://github.com/target/lorri/pull/232#issuecomment-559839797.

Fixes #152:

> Lorri seems to watch directories recursively when it detects `builtins.readDir`. This seems unnecessary because `readDir` only returns the direct children. Any use of those children should be detectable with existing means.

This "minimal diff" fix changes the following behaviour:
- _previously_ when `lorri` saw a `readDir`, it started watching that directory recursively
- _now_ it just watches that directory.

Note that when watching a directory, `notify` will still tell us when any of its children are updated:

> If `recursive_mode` is `RecursiveMode::Recursive` events will be delivered for all files in that tree. **Otherwise only the directory and its immediate children will be watched.**

https://docs.rs/notify/5.0.0-pre.1/notify/trait.Watcher.html#tymethod.watch

We are not using `RecursiveMode::Recursive` - "only the directory and its immediate children" is what we want here.

See https://github.com/target/lorri/issues/152#issuecomment-555505899.

### DIFF
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -69,9 +69,6 @@ impl Watch {
     pub fn extend(&mut self, paths: &[PathBuf]) -> Result<(), notify::Error> {
         for path in paths {
             self.add_path(&path)?;
-            if path.is_dir() {
-                self.add_path_recursively(&path)?;
-            }
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


curiousleo:
PTAL

curiousleo:
PTAL

curiousleo:
PTAL

curiousleo:
I now believe that this is not quite right - the title of this PR as it currently stands is misleading.

I still think that, as the description states, `readDir` should make lorri add a watch just for that directory, non-recursively. However, currently this code can't tell whether the directory came from a `readDir` or a `"copied source"`. In the case of `"copied source"`, I believe that we do need a recursive watch.

hamishmack:
This seems to fix https://github.com/input-output-hk/haskell.nix/issues/492.  Is there a reason it has not been merged?

curiousleo:
@hamishmack does https://github.com/target/lorri/pull/232#issuecomment-559839797 answer your question?

hamishmack:
Yes sorry, not sure how I missed that.

Profpatsch:
Finally fixed by https://github.com/nix-community/lorri/pull/34, PTAL

PR #235, opened 2019-11-21
Add screencast

curiousleo:
Include a GIF version of the asciicast from the announcement blog post in the top-level README.

**[Preview](https://github.com/curiousleo/lorri/blob/include-screencast/README.md#demo)**


curiousleo:
I think the screencast gives potential users a nice idea of what `lorri` is without forcing them to read paragraphs of text. WDYT?

PR #243, opened 2019-11-26
Use slog for logging

### LINKED ISSUE #224: Show journal logs on a single line
Currently lorri logs 10 different log lines during a `lorri direnv` dumping some structure.

When using the NixOS module, logs are collected via the per-user journal.

Showing these logs now looks a bit odd:

```
Nov 15 16:50:20 xps lorri[27661]: Started
Nov 15 16:50:23 xps lorri[27661]: Completed(
Nov 15 16:50:23 xps lorri[27661]:     BuildResults {
Nov 15 16:50:23 xps lorri[27661]:         output_paths: OutputPaths {
Nov 15 16:50:23 xps lorri[27661]:             shell_gc_root: RootPath(
Nov 15 16:50:23 xps lorri[27661]:                 "/home/flokli/.cache/lorri/gc_roots/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/gc_root/shell_gc_root",
Nov 15 16:50:23 xps lorri[27661]:             ),
Nov 15 16:50:23 xps lorri[27661]:         },
Nov 15 16:50:23 xps lorri[27661]:     },
Nov 15 16:50:23 xps lorri[27661]: )
```

Can we just log JSON, and not pretty print it, at least when logging to the journal?

curiousleo:
Fixes #224.

### Motivation

We currently mix logging calls (`debug!`, `info!`, `warn!`, `error!`) with the "print to console" calls `println!` (to stdout) and `eprintln!` (to stderr). That leads to inconsistent-looking output. All feedback to the user should go through logging calls with an appropriate priority.

The `slog` family of libraries supports lots of backends, including pushing logs directly to systemd. So using `slog` opens up possibilities for supporting more backends in the future.

### What's in this PR?

- Removes all `println!` and `eprintln!` calls from non-test code (except from `src/ops/upgrade/mod.rs`, which is currently being reworked in a separate PR) and replaces them with logging calls, making their priority explicit
- Replaces `log` and `env-logger` libraries with libraries from the `slog` ecosystem
- Uses `slog_scope` to inject thread-local logger instances into subcommand calls, reducing boilerplate (passing around loggers gets old quickly)

### Example output

<pre>$ lorri info
Nov 27 17:14:46.820 <font color="#4E9A06">INFO</font> <b>Hello Nix!</b>, <b>root</b>: /home/leo/Code/lorri/shell.nix, <b>lorri_version</b>: 1
</pre>

<pre>$ lorri daemon
Nov 27 18:07:10.073 <font color="#4E9A06">INFO</font> <b>ready</b>, <b>lorri_version</b>: 1
Nov 27 18:07:11.969 <font color="#4E9A06">INFO</font> <b>received ping</b>, <b>nix_file</b>: /home/leo/Code/lorri/shell.nix, <b>lorri_version</b>: 1
Nov 27 18:07:11.970 <font color="#4E9A06">INFO</font> <b>build status</b>, <b>message</b>: Started(ProjectAdded(NixFile(&quot;/home/leo/Code/lorri/shell.nix&quot;))), <b>lorri_version</b>: 1
Nov 27 18:07:15.875 <font color="#4E9A06">INFO</font> <b>build status</b>, <b>message</b>: Completed(BuildResults { output_paths: OutputPaths { shell_gc_root: RootPath(&quot;/home/leo/.cache/lorri/gc_roots/2043910c8519bf079838568f54a73839/gc_root/shell_gc_root&quot;) } }), <b>lorri_version</b>: 1</pre>

### Possible follow-ups

- Push logs directly to journald using `slog_journald`
- Format complex values using the ["nested-values" feature](https://github.com/slog-rs/slog/blob/master/CHANGELOG.md#added-1)


src/ops/daemon.rs
```diff
@@ -49,14 +50,14 @@ pub fn main() -> OpResult {
     })
     .expect("Failed to spawn accept-loop");
 
-    pool.spawn("build-loop", || {
+    pool.spawn("build-loop", move || {
         for msg in build_messages_rx {
-            println!("{:#?}", msg);
+            info!("build status"; "message" => format!("{:?}", msg));
```
curiousleo: This is, strictly speaking, the fix for #224.

curiousleo:
PTAL

Profpatsch:
I merged the update PR, so there’s a few conflicts now.

curiousleo:
> I merged the update PR, so there’s a few conflicts now.

Thanks! I've resolved the conflicts.

curiousleo:
PTAL

src/logging.rs
```diff
@@ -1,49 +1,26 @@
...
+        // direnv swallows stdout, so we must log to stderr
+        Command::Direnv(_) => slog_term::TermDecorator::new().stderr().build(),
+        _ => slog_term::TermDecorator::new().stdout().build(),
+    };
+    let drain = slog_term::FullFormat::new(decorator)
+        .build()
+        .filter_level(level)
+        .fuse();
+    let drain = slog_async::Async::new(drain)
+        .overflow_strategy(slog_async::OverflowStrategy::Block)
```
curiousleo: @Profpatsch I after we saw messages being dropped yesterday. This way messages will never get dropped.

src/main.rs
```diff
@@ -72,27 +77,33 @@ fn create_project(paths: &constants::Paths, shell_nix: NixFile) -> Result<Projec
...
+        Command::Ping_(opts) => Some(&opts.nix_file),
+        _ => None,
+    };
+    let (log, project) = if nix_file.is_none() {
+        (log, None)
+    } else {
+        let project = get_shell_nix(nix_file.unwrap()).and_then(|sn| create_project(&paths, sn))?;
+        let log = log.new(slog::o!("root" => project.nix_file.clone()));
+        (log, Some(project))
+    };
```
grahamc: What would you think about this alternative take on this code's path:

1. adding to cli.rs:


```rust
impl Command {
    pub fn nix_file(&self) -> Option<&PathBuf> {
        match self {
            Command::Info(opts) => Some(&opts.nix_file),
            Command::Direnv(opts) => Some(&opts.nix_file),
            Command::Watch(opts) => Some(&opts.nix_file),
            Command::Ping_(opts) => Some(&opts.nix_file),
            _ => None,
        }
    }
}
```

and then this part of this function becomes:

```rust
    let paths = lorri::ops::get_paths()?;

    let project: Option<Project> = opts
        .command
        .nix_file()
        .map(|nix_file: &PathBuf| get_shell_nix(nix_file))
        .transpose()?
        .map(|nix_file: NixFile| create_project(&paths, nix_file))
        .transpose()?;

    let logged_root: Option<NixFile> = project.as_ref().map(|p| p.nix_file.clone());

    let log = log.new(slog::o!("root" => logged_root));
```

and as a side-effect, I think the following code can be improved a bit too -- removing `unwrap`s and what-not.

src/ops/direnv/mod.rs
```diff
@@ -38,48 +39,54 @@ pub fn main(project: Project) -> OpResult {
...
 
     if std::env::var("DIRENV_IN_ENVRC") != Ok(String::from("1")) {
-        eprintln!(
-            "Warning: 'lorri direnv' should be executed by direnv from within an `.envrc` file."
-        )
+        warn!("`lorri direnv` should be executed by direnv from within an `.envrc` file")
     }
 
-    ok_msg(format!(
+    // direnv interprets stdout as a script that it evaluates. That is why (1) the logger for
```
grahamc: I'm not sure about this change. I quite liked that the caller dealt with what to do with the result, instead of having to pass in different writers. If the caller is already capable of passing in the correct writer, it seems the caller is already capable of dealing with the return value, no?

src/ops/info.rs
```diff
@@ -1,17 +1,13 @@
...
-    println!("lorri version: {}", VERSION_BUILD_REV);
-    println!("Lorri Project Configuration");
-    println!();
-
-    println!("expression: {}", project.nix_file);
-
+pub fn main() -> OpResult {
+    // lorri version and root shell.nix file are attached to the logger as key-value pairs, so they
+    // are part of every log line.
+    info!("Hello Nix!");
```
grahamc: Let's revert this change -- the `println` output, I think, is useful as part of UX for the user. If we decide to drop this op, let's delete the whole op... but not sure we should right now.

src/ops/init.rs
```diff
@@ -1,19 +1,20 @@
...
+use slog_scope::info;
 use std::fs::File;
 use std::io;
 use std::io::Write;
 use std::path::Path;
 
 fn create_if_missing(path: &Path, contents: &str, msg: &str) -> Result<(), io::Error> {
     if path.exists() {
-        println!("- {} {}", msg, path.display());
+        info!("skipped file"; "path" => path.to_str(), "message" => msg);
```
grahamc:
```suggestion
        info!("file already exists, skipping"; "path" => path.to_str(), "message" => msg);
```

What would you think about this?

src/ops/info.rs
curiousleo: Reverted the change.

src/ops/direnv/mod.rs
curiousleo: This way the caller still decides what to do with the output, and arguably it's more explicit (upside in my book) but also more verbose (downside). The way the diff currently also just makes it as unclear as previously what should actually be done with the result. I'll have another think about this and perhaps try to make it closer to what it was before.

curiousleo: I've pushed a different proposal that changes `OpResult` to be `Result<(), ExitError>` after noticing that `ok_msg` was hardly used. I think this makes the code easier to read.

src/main.rs
curiousleo: Thanks for the suggestion! This is a big improvement in readability. I wonder if we can go further. I'll have a think about this.

curiousleo: I've pushed a change that uses more explicit setup code in `main.rs`. It's not the prettiest code in the world, but its purpose is reasonably clear I believe. Let me know what you think!

curiousleo:
PTAL

curiousleo:
PTAL

PR #253, opened 2019-12-09
Use varlink for internal RPCs

curiousleo:
This PR replaces lorri's own RPC library with varlink. I believe that lorri's RPC needs are well served by an external library, and this relieves us from having to maintain one ourselves. Pulling in an additional external library is not without cost, but I believe the trade-off to be in favour in this case.

**Why `varlink`?**
I did a little survey of the Rust RPC landscape. My criteria were:
- Unix socket support
- Streaming RPC support
- Not too heavy

Here's what I found:

|library|supports Unix sockets|supports streaming RPCs|comment|
|---|---|---|---|
|[tonic](https://github.com/hyperium/tonic)|:red_circle: https://github.com/hyperium/tonic/issues/136|:heavy_check_mark:|-|
|[grpcio](https://github.com/pingcap/grpc-rs)|:red_circle: https://github.com/tikv/grpc-rs/issues/296|:heavy_check_mark:|-|
|[tarpc](https://github.com/google/tarpc)|:red_circle: https://github.com/google/tarpc/issues/234|:red_circle: https://github.com/google/tarpc/issues/189|-|
|[capnproto](https://github.com/capnproto/capnproto-rust)|:heavy_check_mark:|:heavy_check_mark:|quite heavyweight for our purposes|
|[varlink](https://github.com/varlink/rust/)|:heavy_check_mark:|:heavy_check_mark:|fits the bill|

As a result of this little survey, I went ahead with varlink. Note that there are less than 30 lines of varlink-specific code and the varlink interface definition looks similar to other interface definition languages, so we can change the RPC framework in the future if we find one that better suits our needs.

**Todo:**

- [x] Remove `thread::sleep` from `tests/daemon/main.rs`
- [x] Remove `unwrap` calls where error handling should be more explicit
- [x] Define errors, return to `ping_`
- [x] Explain motivation in more detail
- [x] Make CI pass

Currently `nix-build` fails on CI because of a Carnix issue. I've filed https://nest.pijul.com/pmeunier/carnix/discussions/41 and opened https://github.com/varlink/rust/pull/32. In the meantime, I've forked `varlink_derive` and made a one-line change to make Carnix happy.


tests/daemon/main.rs
```diff
@@ -82,28 +77,7 @@ pub fn start_job_with_ping() -> std::io::Result<()> {
...
     }?;
 
+    drop(accept_handle);
     drop(tempdir);
-    daemon_subroutine_handle.join().unwrap();
-    Ok(())
-}
-
-#[test]
-pub fn start_two_listeners_on_same_socket() -> std::io::Result<()> {
```
curiousleo: This test has been replaced by `lock_is_exclusive` in `socket.rs`.

curiousleo:
Currently `nix-build` fails on CI because of a Carnix issue. I've filed https://nest.pijul.com/pmeunier/carnix/discussions/41 and opened https://github.com/varlink/rust/pull/32 to make progress.

src/daemon/rpc.rs
```diff
@@ -0,0 +1,70 @@
...
+        let service = varlink::VarlinkService::new(
+            /* vendor */ "com.target",
+            /* product */ "lorri",
+            /* version */ "0.1",
+            /* url */ "https://github.com/target/lorri",
+            vec![Box::new(rpc::new(Box::new(self)))],
+        );
+        varlink::listen(
+            service, address, /* initial_worker_threads */ 1, /* max_worker_threads */ 1,
+            /* idle_timeout */ 0,
```
grahamc: may I suggest instead:

```rust
let initial_worker_threads = 1;
let max_worker_threads = 1;
let idle_timeout = 0;
varlink::listen(initial_worker_threads, max_worker_threads, idle_timeout);
```

src/com.target.lorri.varlink
```diff
@@ -0,0 +1,53 @@
...
+
+# WatchShell instructs the daemon to evaluate a Nix expression and re-evaluate
+# it when it or its dependencies change.
+method WatchShell(shell_nix: ShellNix) -> ()
+
+# ShellNix describes the Nix expression which evaluates to a development
+# environment.
+type ShellNix (
+  # The absolute path of a Nix file specifying the project environment.
+  path: string
```
grahamc: I wonder if we can somehow indicate to varlink that this must be converted in to a Path/PathBuf, to avoid conversions

src/ops/direnv/mod.rs
```diff
@@ -16,20 +15,19 @@ use std::process::Command;
 pub fn main<W: std::io::Write>(project: Project, mut shell_output: W) -> OpResult {
     check_direnv_version()?;
 
-    let socket_path = crate::ops::get_paths()?.daemon_socket_file().to_owned();
-
     let root_paths = Roots::from_project(&project).paths();
     let paths_are_cached: bool = root_paths.all_exist();
+    let address = crate::ops::get_paths()?.daemon_socket_address();
+    let shell_nix = rpc::ShellNix {
```
grahamc: What if we implemented a TryFrom for ShellNix <-> NixFile? Then we don't need to implement the error checking and conversion time multiple times.

src/com.target.lorri.varlink
curiousleo: I don't think the Varlink code generation can be modified in that way without major hacks, but https://github.com/target/lorri/pull/253#discussion_r356688072 works pretty well!

src/ops/direnv/mod.rs
curiousleo: Done, nice!

src/daemon/rpc.rs
curiousleo: Yeah, this looked better in a previous iteration when the arguments were all aligned. I've implemented your suggestion now.

src/com.target.lorri.varlink
curiousleo: For reference, here is the list of Varlink-supported types: https://varlink.org/Interface-Definition

curiousleo:
For reference: `varlink_generator` (with the checked-in `com_target_lorri.rs`) is used here instead of the more elegant `varlink_derive` (which would have generated that module on the fly) because with `varlink_derive`, the Nix build failed as follows:
```
error: extern location for varlink_derive does not exist: /nix/store/khk2k1xkb8yax0bidg19pfp0kw44x9v9-rust_varlink_derive-9.0.0-lib/lib/libvarlink_derive-015e3fde5f.rlib
```

I don't know why. I assume this has _something_ to do with Carnix because `cargo build` is happy with `varlink_derive`.

PR #254, opened 2019-12-10
Use Rust 1.39.0 in CI

curiousleo:
Rust 1.39.0 is the latest stable version of Rust. Crucially, `await` and `async` shipped in this version. They are prerequisites for `lorri services`.

**What's in this PR?**
- Updated `nixpkgs.json` to the latest `nixos-unstable`
- Replaced uses of the "stable" Mozilla overlay channel by `nixpkgs.rust`, which is at 1.39.0 in `nixos-unstable`
- Replaced `racer.nix` by `nixpkgs.rustracer`; as a result, we don't need Rust nightly anymore
- Removed the Mozilla overlay because it is no longer necessary

I think this simplifies the setup quite a bit. One potential downside is that we previously built racer 2.1.28 from source, whereas now we use `nixpkgs.rustracer` which is at 2.1.27. But as far as I can tell, there were [no functional changes](https://github.com/racer-rust/racer/compare/v2.1.27...v2.1.28) in the 2.1.28 release.


curiousleo:
Todo: update nixos-unstable channel; use pkgs.rust_1_39 from there for the CI build.

curiousleo:
@Profpatsch you may be interested in this too.

PR #256, opened 2019-12-13
logged-evaluation.nix: support additional servicesSrc argument

curiousleo:
**Context:** we're in the process of adding support for `services.nix`, a nix file specifying long-running services that enable / support project development.

This PR contains 100% plumbing. There are zero user facing changes.

In order to evaluate `services.nix` without too much additional overhead, the idea is to coalesce, or batch, the evaluation. That is, if there is both a `shell.nix` and a `services.nix`, they should be evaluated by lorri using a single `nix-instantiate` subprocess whenever necessary.

This PR adds support for a `servicesSrc` argument to `logged-evaluation.nix`. If this argument is used, evaluating `logged-evaluation.nix` will result in two derivations, one for the shell and one for services. This is a first step towards supporting `services.nix` in addition to `shell.nix`.

**Why not split this already large file into multiple files?** I've decided to keep it all in `logged-evaluation.nix` because every additional file would have to be included in the Rust binary using its own `include_str!` (at least the way things are currently set up). This doesn't add all that much to `logged-evaluation.nix` so that seemed fine.

### DIFF
--- a/src/logged-evaluation.nix
+++ b/src/logged-evaluation.nix
@@ -1,24 +1,24 @@
-{ src, runTimeClosure }:
+{ shellSrc, servicesSrc ? null, runtimeClosure }:
 let
-  runtimeCfg = import runTimeClosure;
+  runtimeCfg = import runtimeClosure;
 
   # using scopedImport, replace readDir and readFile with
   # implementations which will log files and paths they see.
-  overrides = {
-    import = scopedImport overrides;
-    scopedImport = x: builtins.scopedImport (overrides // x);
-    builtins = builtins // {
-      readFile = file: builtins.trace "lorri read: '${toString file}'" (builtins.readFile file);
-      readDir = path: builtins.trace "lorri read: '${toString path}'" (builtins.readDir path);
-    };
-  };
-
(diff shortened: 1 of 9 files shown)


curiousleo:
Merge conflict alert: this clashes with https://github.com/target/lorri/pull/253. Let's merge #253 first. I'm working on a branch rebased on top of #253 in the meantime.

This PR is good to be reviewed, but should not be merged.

src/logged-evaluation.nix
```diff
@@ -31,7 +31,7 @@ let
   # gc rooting the resulting store path from this build will retain
   # references to all the store paths needed, preventing the shell's
   # actual environment from being deleted.
-  keep-env-hack = drv: derivation (drv.drvAttrs // {
+  wrapped-shell = drv: derivation (drv.drvAttrs // {
     name = "lorri-keep-env-hack-${drv.name}";
```
grahamc: Let's change the name here to be shell-specific, too.

curiousleo:
Closing in favour of https://github.com/target/lorri/pull/257.

PR #257, opened 2019-12-16
Generate services.json

curiousleo:
**Motivation:** we want to be able to generate a `services.json` file based on a nix file. This `services.json` is meant to be consumed by the `lorri services` client to start and stop development services.

On the Nix side, this makes both `shellSrc` and `servicesSrc` optional; at least one of them must be provided.

On the Rust side, `NixFile` is now an enum with possible values `NixFile::Shell(PathBuf)` and `NixFile::Services(PathBuf)`.

**Todo**
- [x] Test `builder::run` with `NixFile::Services`
- [x] Check nix file formatting in CI

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,13 +75,6 @@ impl From<&NixFile> for PathBuf {
     }
 }
 
-/// Proxy through the `Display` class for `PathBuf`.
-impl std::fmt::Display for NixFile {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        PathBuf::from(self).display().fmt(f)
-    }
-}
-
 impl slog::Value for NixFile {
     fn serialize(
         &self,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,31 +59,26 @@ include!(concat!(env!("OUT_DIR"), "/build_rev.rs"));
 
 /// A .nix file.
 #[derive(Hash, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
-pub struct NixFile(PathBuf);
+pub enum NixFile {
+    /// A .nix file which describes a shell environment
+    Shell(PathBuf),
+    /// A .nix file which describes a list of services
+    Services(PathBuf),
+}
 
-impl NixFile {
-    /// Underlying `&OsStr`.
-    pub fn as_os_str(&self) -> &std::ffi::OsStr {
-        self.0.as_os_str()
+impl From<&NixFile> for PathBuf {
+    fn from(p: &NixFile) -> PathBuf {
+        match p {
+            NixFile::Shell(p) => p.to_path_buf(),
+            NixFile::Services(p) => p.to_path_buf(),
+        }
     }
 }
 
 /// Proxy through the `Display` class for `PathBuf`.
 impl std::fmt::Display for NixFile {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        self.0.display().fmt(f)
-    }
-}
-
-impl From<&std::ffi::OsStr> for NixFile {
-    fn from(s: &std::ffi::OsStr) -> NixFile {
-        NixFile(PathBuf::from(s.to_owned()))
-    }
-}
-
-impl From<PathBuf> for NixFile {
-    fn from(p: PathBuf) -> NixFile {
-        NixFile(p)
+        PathBuf::from(self).display().fmt(f)
     }
 }
 
@@ -94,7 +89,7 @@ impl slog::Value for NixFile {
         key: slog::Key,
         serializer: &mut dyn slog::Serializer,
     ) -> slog::Result {
-        serializer.emit_str(key, &self.as_os_str().to_string_lossy())
+        serializer.emit_arguments(key, &format_args!("{}", PathBuf::from(self).display()))
     }
 }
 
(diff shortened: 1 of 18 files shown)


src/logged-evaluation.nix
```diff
@@ -1,20 +1,19 @@
-{ src, runTimeClosure }:
+{ shellSrc, servicesSrc ? null, runtimeClosure }:
```
grahamc: How about adding a few comments to the function declaration, describing the types of these values a bit?

src/logged-evaluation.nix
```diff
@@ -16,9 +16,10 @@ let
   imported =
```
grahamc: Let's get nixpkgs-fmt to be part of CI, and fail the build if any `*.nix` file isn't formatted properly. (`find . -name '*.nix' | xargs nixpkgs-fmt` I can help more with this)

src/logged-evaluation.nix
curiousleo: I've commented the function and the `shellSrc` and `servicesSrc` arguments. What do you think about how this looks now?

src/logged-evaluation.nix
curiousleo: I've added `nix/fmt.sh` which does essentially this, but excludes generated nix files.

src/logged-evaluation.nix
```diff
@@ -31,30 +31,30 @@ let
   # gc rooting the resulting store path from this build will retain
   # references to all the store paths needed, preventing the shell's
   # actual environment from being deleted.
-  wrapped-project = drv: services: derivation (
-    drv.drvAttrs // {
-      name = "lorri-wrapped-project-${drv.name}";
+  wrapped-project = shell: services: derivation (
```
Profpatsch: Nitpick: this can be an arbitrary `drv`, so naming it `drv` does make sense.

src/logged-evaluation.nix
```diff
@@ -32,16 +33,16 @@ let
...
 
       origExtraClosure = shell.extraClosure or [];
       extraClosure = runtimeCfg.closure;
 
-      origBuilder = shell.builder;
+      origBuilder = shell.builder or null;
       builder = runtimeCfg.builder;
 
-      origSystem = shell.system;
+      origSystem = shell.system or null;
```
Profpatsch: Having to introduce so many optional fields means we have a leaky abstraction.

src/lib.rs
```diff
@@ -59,31 +59,26 @@ include!(concat!(env!("OUT_DIR"), "/build_rev.rs"));
 
 /// A .nix file.
 #[derive(Hash, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
-pub struct NixFile(PathBuf);
+pub enum NixFile {
+    /// A .nix file which describes a shell environment
+    Shell(PathBuf),
+    /// A .nix file which describes a list of services
+    Services(PathBuf),
+}
```
Profpatsch: This doesn’t look like the right enum to put this in, rather there should be a second enum with two `NixFiles` in the fields.

src/lib.rs
```diff
@@ -75,13 +75,6 @@ impl From<&NixFile> for PathBuf {
     }
 }
 
-/// Proxy through the `Display` class for `PathBuf`.
-impl std::fmt::Display for NixFile {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        PathBuf::from(self).display().fmt(f)
-    }
-}
```
Profpatsch: This is a symptom of adding another field to `NixFile`, even though it should be a separate enum.

src/main.rs
```diff
@@ -88,8 +88,8 @@ fn run_command(log: slog::Logger, opts: Arguments) -> OpResult {
 
     match opts.command {
         Command::Info(opts) => {
-            let (project, _guard) = with_project(&opts.nix_file)?;
-            info::main(project)
+            let (_project, _guard) = with_project(&opts.nix_file)?;
```
Profpatsch: Not using any of the fields, so we can remove this line.

src/logged-evaluation.nix
curiousleo: While it can be an arbitrary `drv`, in the context in which this is used, we'll always have something representing a shell and something representing services, so I think those names are preferable because the communicate the intention better.

src/lib.rs
curiousleo: The way I thought about this is simply that lorri deals with two types of nix files, namely ones that describe a shell and ones that describe services. The `NixFile` enum describes exactly that.

src/main.rs
curiousleo: The `_guard` is required. I believe that calling things `_guard` is something of a convention which indicates that the variable, though unused, must stay in scope.

In this case, the `_guard` protects the thread-local logger from being destroyed.

`_project` is indeed unused.

Would you prefer the following?

```rust
let _guard = with_project(&opts.nix_file)?.1;
```

src/logged-evaluation.nix
curiousleo: I've attempted to address this here: https://github.com/target/lorri/pull/266

src/main.rs
Profpatsch: Oh, right, we need it for scoping. Then it’s fine. I didn’t know about the `_guard` convention.

src/lib.rs
Profpatsch: I guess we can go both ways.

PR #259, opened 2019-12-18
lorri services with own build loop

curiousleo:
This is an MVP of the `lorri services` command.  
  
**What does it do?**  
`lorri services` is a blocking command. It takes a `--services` argument, which defaults to `services.nix`. This nix file describes a list of services. `lorri services` starts them initially. It then watches the services nix file and its dependencies for changes and rebuilds the list of services. Then it starts the newly added services, stops those that were removed, and restarts those whose configuration has changed.  
  
**In what sense is this an MVP?**  
  
* Services that terminate on their own are not restarted. Restarting services is a tricky business - see the large number of service restarting related options in systemd.      
  
* Building the services is done by `lorri services` itself. Down the line, this should be done by the daemon to deduplicate builds. As an example, if `shell.nix` and `services.nix` share a dependency and this dependency changes, the environment and the services will be rebuilt separately in the MVP. They could be rebuilt in a single call to nix if the daemon was responsible for building services as well as environment.      

**Todo**
- [x] Add unit test
- [x] Add integration test
- [x] Make this `services.nix` work:
  ```nix
  let pkgs = import <nixpkgs> {}; in [
    { name = "httpd"; program = "${pkgs.python3}/bin/python3"; args = [ "-m" "http.server" ]; }
    { name = "notify"; program = "${pkgs.bash}/bin/bash"; args = [ "-c" "ls | ${pkgs.entr}/bin/entr echo \"file changed:\" /_" ]; }
  ]


curiousleo:
Current issue:
```
"error: in \'toFile\': the file \'lorri-wrapped-project\' cannot refer to derivation outputs, at /home/leo/.cache/lorri/cas/fbe1103295e2a953e7f81ff7b4c5cef7:126:11"
```

Using
```nix
let pkgs = import <nixpkgs> {}; in [
  { name = "httpd"; program = "${pkgs.python3}/bin/python3"; args = [ "-m" "http.server" ]; }
  { name = "notify"; program = "${pkgs.bash}/bin/bash"; args = [ "-c" "ls | ${pkgs.entr}/bin/entr echo \"file changed:\" /_" ]; }
]
```

curiousleo:
I looks like using tokio, or any other threaded futures executor, will not be so easy. The problem is that `tokio` with the `rt-threaded` feature [1] as well as `futures` with the `thread-pool` feature [2] require `num_cpus` which requires `hermit-abi` [3] when the OS is `hermit`.

`hermit-abi` itself does not build with Rust stable due to its use of `!#[feature]`.

This is a problem because carnix does not understand conditional dependencies and builds `hermit-abi` even though we don't care about hermit OS.

[1] https://github.com/tokio-rs/tokio/blob/tokio-0.2.8/tokio/Cargo.toml#L73
[2] https://github.com/rust-lang/futures-rs/blob/0.3.1/futures-executor/Cargo.toml#L20
[3] https://github.com/seanmonstar/num_cpus/blob/v1.11.1/Cargo.toml#L17-L18
[4] https://github.com/hermitcore/rusty-hermit/blob/master/hermit-abi/src/lib.rs#L1

curiousleo:
I'm trying to figure out why the new integration test is failing.

`strace` of a successful run:
> [pid 32044] stat("/run/user/1000/.tmpc5Tzy2/touchit", 0x7f589fa10f20) = -1 ENOENT (No such file or directory)
> [...]
> [pid 32106] execve("/nix/store/j0lhk5xkky410h5v9af98sz6kd8cbp40-coreutils-8.31/bin/touch", ["touch", "/run/user/1000/.tmpc5Tzy2/touchi"...], 0x7ffe743328c8 /* 144 vars */ <unfinished ...>

`strace` of an unsuccessful run (from CI):
> [pid 16166] stat("/run/user/2000/.tmpXcsI2Q/touchit", 0x7fcec22edf20) = -1 ENOENT (No such file or directory)^M
> [...]
> [pid 16196] execve("/nix/store/v826z0g5maw6hvgm5595jr9x5djx4cfw-lorri-runtime-tools/bin/touch", ["touch", "/nix/store/xvg60fs4xs041mh65dkhz"...], 0x4e3008 /* 30 vars */) = 0^M

It's weird that in the unsuccessful run, `touch` seems to be called with a path inside `/nix/store` as an argument instead of "/run/user/2000/.tmpXcsI2Q/touchit", the file in the temporary directory that it is supposed to touch.

curiousleo:
Next step: enable logging in the integration test to figure out where things fail.

Profpatsch:
I think we can close this, since the research money has run out.

Ideally we’d have a tool like `lorri services` in the future, but it wasn’t meant to be.

PR #265, opened 2020-01-06
Allow command failures in shell hook

### LINKED ISSUE #264: Error in lorri-keep-env-hack-*.drv when shellHook returns non-zero
The following `shell.nix` **works with nix-shell** but doesn't work with lorri for the reason stated in the title above.

```nix
(import <nixpkgs> {}).stdenv.mkDerivation {
  name = "shell";
  shellHook = ''
    echo foo
    $(exit 1)
  '';
}
```

`lorri watch` yields:

```
[...]
            "these derivations will be built:",
            "  /nix/store/bq4f2as76d9hxvsz7afghn27n7lx8fwk-lorri-keep-env-hack-shell.drv",
            "building \'/nix/store/bq4f2as76d9hxvsz7afghn27n7lx8fwk-lorri-keep-env-hack-shell.drv\'...",
            "foo",
            "builder for \'/nix/store/bq4f2as76d9hxvsz7afghn27n7lx8fwk-lorri-keep-env-hack-shell.drv\' failed with exit code 1",
            "error: build of \'/nix/store/bq4f2as76d9hxvsz7afghn27n7lx8fwk-lorri-keep-env-hack-shell.drv\' failed",
        ],
    },
)
```

curiousleo:
Closes #264.

This mirrors the behaviour of nix-build; see https://github.com/NixOS/nix/blob/2.3.1/src/nix-build/nix-build.cc#L422.

### DIFF
--- a/src/logged-evaluation.nix
+++ b/src/logged-evaluation.nix
@@ -137,9 +137,11 @@ let
 
             # target/lorri#23
             # https://github.com/NixOS/nix/blob/bfc6bdf222d00d3cb1b0e168a5d55d1a7c9cdb72/src/nix-build/nix-build.cc#L424
+            set +e
             if [ "$(type -t runHook)" = function ]; then
              runHook shellHook;
             fi;
+            set -e
 
             export > "$out/bash-export"
             cat << 'EOF' > "$out/services.json"
--- a/src/logged-evaluation.nix
+++ b/src/logged-evaluation.nix
@@ -132,15 +132,17 @@ let
 
             export IN_NIX_SHELL=impure
 
-            # https://github.com/NixOS/nix/blob/92d08c02c84be34ec0df56ed718526c382845d1a/src/nix-build/nix-build.cc#
+            # https://github.com/NixOS/nix/blob/2.3.1/src/nix-build/nix-build.cc#L418
             [ -e $stdenv/setup ] && . $stdenv/setup
 
-            # target/lorri#23
-            # https://github.com/NixOS/nix/blob/bfc6bdf222d00d3cb1b0e168a5d55d1a7c9cdb72/src/nix-build/nix-build.cc#L424
+            # https://github.com/NixOS/nix/blob/2.3.1/src/nix-build/nix-build.cc#L422
             set +e
+
+            # https://github.com/NixOS/nix/blob/2.3.1/src/nix-build/nix-build.cc#L424
             if [ "$(type -t runHook)" = function ]; then
              runHook shellHook;
             fi;
+
             set -e
 
             export > "$out/bash-export"


src/logged-evaluation.nix
```diff
@@ -132,15 +132,19 @@ let
 
             export IN_NIX_SHELL=impure
 
-            # https://github.com/NixOS/nix/blob/92d08c02c84be34ec0df56ed718526c382845d1a/src/nix-build/nix-build.cc#
+            # https://github.com/NixOS/nix/blob/2.3.1/src/nix-build/nix-build.cc#L418
```
grahamc: I have a fairly weak preference towards using actual commits instead of tags -- and optionally adding a `(tag: 2.3.1)` at the end. In this URL it is a bit unclear if 2.3.1 is a branch or a tag itself, with a bit of added worry about tags being technically mutable. Same goes for the other URLs.

curiousleo:
> I'm not sure about this: the shellHook might be failing for very important and dangerous reasons that the user will not know about, and then continue to successfully provide an environment without the user seeing a problem.
> 
> I sort of think the real problem behind #264 is we have no good way to communicate to the user that their shell.nix isn't working. What do you think?

My thoughts on this are:
- I can't really think of an example `shellHook` where a command fails and I as a user would not want to know about it - but it seems like https://github.com/target/lorri/issues/264 was based on a real world example of exactly this.
- I think that it is a reasonable user expectation that `lorri` is a drop-in for `nix-shell`. We should be explicit about when that is not the case.
- Having things fail silently is generally a bad thing.

So I think there's a trade-off here:
- We can try to make lorri a `nix-shell` drop-in replacement as much as possible. That means trying to support most `shell.nix` in the wild, possibly including ones where shellHook commands fail. That's the approach taken in this PR.
- Or we think of lorri as more of a thing on its own, where we can impose our own (sensibly argued) restrictions (and offer extensions) on what people can put in the `shell.nix`. That's the main alternative approach I can think of, and in that case we should (1) make lorri fail with a helpful error message when a command in shellHook fails (though this may be difficult given that lorri just observes nix evaluate the file), and (2) document that lorri deviates from nix-shell behaviour in this case.

> _also: this is actually not about pipeline failures (that would be `set [-+]o pipefail`) but about regular commands failing_

Thanks for the clarification! I've changed the PR title, and will change the commit message too.

src/logged-evaluation.nix
curiousleo: Thanks - I'll address this once we've figured out the more high level issues with this PR.

curiousleo:
> In that case we should (1) make lorri fail with a helpful error message when a command in shellHook fails (though this may be difficult given that lorri just observes nix evaluate the file) [...]

This does indeed look difficult. lorri calls nix as an external process and parses its output. In the failure case, lorri already outputs everything the nix process has printed. So lorri doesn't really have any more context to add than is already given to the user. The only thing I can think of is that lorri could point out a list of common causes for errors. That's a lot of guesswork at the moment though since we're still discovering failure modes, so I'm not sure if that would really improve the user experience in the current state of the project.

curiousleo:
Okay, here's an alternative plan suggested by @grahamc:

- We won't `set +e`. If there is a command in the shellHook that fails, then building the environment should fail too.
- We'll execute the shellHook in a separate shell and capture the output. If the shellHook fails (e.g. due to a failing command), we'll point the user to the output.

curiousleo:
> We'll execute the shellHook in a separate shell and capture the output. If the shellHook fails (e.g. due to a failing command), we'll point the user to the output.

Here is how this could work:
- Generate a random file path (possibly inside a temporary directory) and pass it from `instrumented_instantiation` as argstr to `logged_evalution.nix`
- `logged_evaluation.nix` captures the stderr (and possibly stdout) of `runHook shellHook` and writes to the files whose paths have been passed to it from `instrumented_instantiation` in case there is a failure while executing `runHook shellHook`
- If the stderr (or stdout) exist, we know that a failure happened. `instrumented_instantiation` returns the paths of the log files to `run`; `run` returns them to `BuildLoop::once`

The issue is that building is asynchronous. `lorri direnv` only starts the build, but doesn't wait for it to end. So we need to keep track of the fact that the shellHook failed somehow, and report it to the user at the next opportunity.

One option for this would be:

- keep track of shell_file -> last result in daemon
- next invocation: look up last result for the given shell_file
- return last result to `lorri direnv` via `watch_shell` rpc
- print last result to direnv so user can see it

Alternatively, we could use per-nix-file "well-known locations" for the stderr and stdout, i.e. some way for `lorri direnv` to figure by itself if there were shellHook logs for a particular shell.nix file. In that case it could report them to the user directly, without the daemon having to store a mapping from nix files to the latest shellHook logs for that nix file.

curiousleo:
As explained above, we don't want to swallow shellHook failures the way this PR implements it. So I'm closing this PR.

I've created https://github.com/target/lorri/issues/273 to track the issue, with a link to the implementation idea above: https://github.com/target/lorri/pull/265#issuecomment-572949738.

haslersn:
@curiousleo Yes, #264 was based on a real world example:

```
[ -f .nix-shell-hook ] && source .nix-shell-hook
```

This is common syntax.

The first sentence in the lorri description is: "lorri is a nix-shell replacement for project development." With this, I think it is unacceptable to be incompatible in that regard.

curiousleo:
> ```
> [ -f .nix-shell-hook ] && source .nix-shell-hook
> ```

According to the Bash man page, even under `set -e`, this should succeed if either `.nix-shell-hook` does not exist or (it exists and sourcing it succeeds). In other words, this should only _fail_ if `.nix-shell-hook` exists and sourcing it leads to an error.

The question is: would you rather not know that `source .nix-shell-hook` failed? We believe that under most circumstances, it is preferable to know about this failure.

```
-e      [...] The shell does  not  
        exit  if  the  command that fails is [...] part of  any  command  
        executed  in a && or || list except the command following the final && or || [...].
```
https://linux.die.net/man/1/bash

haslersn:
It does not exit bash. It just sets $? to nonzero and lorri interprets this as failure.

PR #271, opened 2020-01-07
Remove unnecessary extern crate declarations

curiousleo:
## Overview

100% refactoring: in Rust 2018, `extern crate` is almost never necessary. So we delete those declarations we don't need.

## Checklist

- [x] Tested the change (unit or integration tests) [project still compiles, tests still pass]

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -324,7 +324,7 @@ fn parse_evaluation_line<T>(line: T) -> LogDatum
 where
     T: AsRef<OsStr>,
 {
-    lazy_static! {
+    lazy_static::lazy_static! {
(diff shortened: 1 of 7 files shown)


src/builder.rs
```diff
@@ -324,7 +324,7 @@ fn parse_evaluation_line<T>(line: T) -> LogDatum
 where
     T: AsRef<OsStr>,
 {
-    lazy_static! {
+    lazy_static::lazy_static! {
```
curiousleo: My heuristic here was: if a macro is only ever used once, then call it directly and remove the `#[macro_use] extern crate` declaration.

Issue #258, opened 2019-12-17
Caching on CI does not seem to work


curiousleo:
Both the nix build and the cargo build appear to always build all dependencies from scratch, which leads to build times between 10 and 30 minutes.

Example: https://travis-ci.org/target/lorri/builds/626109465

Profpatsch:
```
$ nix-build -A allBuildInputs shell.nix > ./shell-inputs
unpacking 'https://github.com/NixOS/nixpkgs/archive/3140fa89c51233397f496f49014f6b23216667c2.tar.gz'...
these derivations will be built:
  /nix/store/kk9c7kkgy344dkfwrbbdizlbi2y2icxh-clippy-1.39.0.drv
these paths will be fetched (447.85 MiB download, 1211.02 MiB unpacked):
  /nix/store/0vcqyg4cadpns1l935h18ram5ngqqsgq-aws-sdk-cpp-1.7.90
  /nix/store/0zgfzj164xhhc56bfgk9q53aqm5hgk9n-patch-2.7.6
```

It’s building clippy on every run, which takes a long time. Maybe that was introduced with the last rust update?

curiousleo:
> Maybe that was introduced with the last rust update?

That's definitely possible. I can see how changing the `clippy` dependency would lead to `clippy` being built _once_ (or once per platform) on Travis. But why keep rebuilding it on every CI run? It looks to me like the "cache" build step is meant to cache cargo outputs: https://github.com/target/lorri/blob/master/.travis.yml.nix#L103

curiousleo:
For nix-based builds, part of the puzzle is of course that we've disabled cachix on macOS due to this issue: https://github.com/cachix/cachix/issues/228#issuecomment-533634704.

Why is `rustPackages.clippy` only available in source form, but not in compiled form, from cache.nixos.org?

Profpatsch:
> For nix-based builds, part of the puzzle is of course that we've disabled cachix on macOS due to this issue: [cachix/cachix#228 (comment)](https://github.com/cachix/cachix/issues/228#issuecomment-533634704).

That’s fixed since #340 

Build times seem to be down to ~8 minutes, which is not rosy, but at least the nix dependencies are cached again.

Profpatsch:
The caching for the macos nix build still has some problems: https://travis-ci.org/target/lorri/jobs/657713960?utm_medium=notification&utm_source=github_status

PR #274, opened 2020-01-10
No clippy on macOS CI

curiousleo:
Partial workaround for https://github.com/target/lorri/issues/258.

## Overview

Cachix does not work on macOS (https://github.com/cachix/cachix/issues/228) and Hydra does not successfully build `clippy` (https://github.com/NixOS/nixpkgs/pull/77327, https://github.com/NixOS/nixpkgs/issues/77358). As a result, in order to run `clippy`, we build it from scratch on every CI run on macOS. This adds about 20 minutes of build time.

There is no need to run `clippy` both on macOS and on Linux. So to circumvent this issue, this PR changes the CI setup slightly so that lints (including `clippy`) are only run on Linux.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ matrix:
     script:
     - set -e
     - nix-build -A allBuildInputs shell.nix > ./shell-inputs
-    - nix-shell --quiet --arg isDevelopmentShell false --run ci_check
+    - nix-shell --quiet --arg isDevelopmentShell false --run ci_test
     - cat $(nix-build --quiet ./.travis.yml.nix --no-out-link) > .travis.yml
     - git diff -q ./.travis.yml
     - git diff -q ./Cargo.nix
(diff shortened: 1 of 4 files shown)


curiousleo:
PTAL. The effect of this is to reduce the time  `nix-build -A allBuildInputs shell.nix > ./shell-inputs` takes on macOS from 1244 seconds (20 minutes) to 95 seconds (1.5 minutes).

- before: https://travis-ci.org/target/lorri/jobs/634125087#L559 (roughly - this is from a different PR, but we don't seem to be running CI on master so this is as good a comparison as I can find right now)
- after: https://travis-ci.org/target/lorri/jobs/635266664#L558

curiousleo:
@Profpatsch @grahamc PTAL

grahamc:
Go ahead and rebase to resolve conflicts, regenerate the .yml and merge.

Issue #269, opened 2020-01-07
Document release process


curiousleo:
https://github.com/target/lorri/blob/master/MAINTAINERS.md#cutting-a-release says:
> TODO

That section should be updated, **answering the following questions:**
- [x] Q: What is the relationship between `master` and `rolling-release`?
  A: `rolling-release` is no longer used; releases are made from `master`.
- [x] Q: How is the version number determined?
  A: we use both revCount and [this versioning scheme documented in MAINTAINERS.md](https://github.com/target/lorri/blob/master/MAINTAINERS.md#versioning-scheme)
- [x] Q: Which distribution channels are supported (currently `nixos-unstable` and `nixos-19.09`)?
  A: `nixos-stable` and `nixos-19.09` are the supported distribution channels.
- [ ] Q: For each channel, what is our update policy (under which circumstances do we update the channel)?
- [x] Q: For each channel, how do we actually release the new lorri version?
  A: See [MAINTAINERS.md](https://github.com/target/lorri/blob/master/MAINTAINERS.md#publishing-a-release-on-nixpkgs)

As an example, lorri `master` may at some point require Rust 1.39 which is not (at this moment) included in `nixos-19.09`, so the `nixos-19.09` update policy may be "backport critical bugfixes" whereas the `nixos-unstable` update policy may be "update every month or so".

curiousleo:
@Profpatsch perhaps you can answer the first question, "What is the relationship between `master` and `rolling-release`" and how `rolling-release` is updated?

![Screenshot from 2020-01-07 13-26-06](https://user-images.githubusercontent.com/697092/71895446-63851700-3151-11ea-9bdc-dbe371912d90.png)

That's something I really don't understand at the moment.

Profpatsch:
> lorri `master` now requires Rust 1.39

Huh, why is that?

> so the `nixos-19.09` update policy may be "backport critical bugfixes" whereas the `nixos-unstable` update policy may be "update every month or so".

That’s the reason we originally didn’t want to put it into nixpkgs, since it’s going to be a considerable maintenance overhead (people opening issues about already fixed bugs, necessary backporting, …).

curiousleo:
> > lorri `master` now requires Rust 1.39
> 
> Huh, why is that?

[...]

PR #281, opened 2020-01-14
Package script

curiousleo:
Relates to #269.

## Overview

This is work towards making it easier to distribute lorri. The goal is to have a script that generates source archives which can be uploaded as releases. We want to be able to build the source in those archives with minimal fuss, e.g. `cargo build` should just work and do the right thing.

```console
$ package=$(nix-build package.nix)
$ cd $(mktemp -d)
$ tar -xzf $package
$ chmod -R u+w src/ # necessary for code generation
$ nix-shell -p cargo rustfmt --run 'cargo build'
$ ./target/debug/lorri info
lorri version: 421
[...]
```

## Checklist


package.nix
```diff
@@ -0,0 +1,73 @@
...
+    tarfile=$(mktemp)
+
+    # Include all files under version control.
+    tar -C $srcPath -cvf $tarfile $(git -C $srcPath ls-files)
+
+    # Include the version, call it "VERSION".
+    tar --transform="s:${drop 1 version}:VERSION:" -rvf $tarfile ${version}
+
+    # Include the runtime closure, call it "RUNTIME_CLOSURE".
+    tar --transform="s:''${closurePath#?}:RUNTIME_CLOSURE:" -rvf $tarfile $closurePath
```
grahamc: Let's hold off on implementing this, the RUNTIME_CLOSURE bit might be quite complicated, and let's focus more on services for the time being. This is because the closure should be calculated by who is building the package.

Profpatsch:
I don’t think we strictly need this, since lorri is nix-specific, so we can expect users to always have a nixpkgs version available, hopefully with a working version of lorri in it.

PR #286, opened 2020-01-16
info command: show gc roots

curiousleo:
## Overview

The `info` command currently shows only information that is accessible via other means too. Let's make it more useful!

In this PR:
- every log line: remove "version" field
- every log line: change field name of the shell nix "root" to "expr"
- info command: keep printing lorri version
- info command: include information about GC roots

## Checklist

### DIFF
--- a/src/ops/info.rs
+++ b/src/ops/info.rs
@@ -8,6 +8,7 @@ use slog_scope::info;
 /// See the documentation for lorri::cli::Command::Info for more
 /// details.
 pub fn main(project: Project) -> OpResult {
+    info!("lorri version: {}", crate::VERSION_BUILD_REV);
     let root_paths = Roots::from_project(&project).paths();
     let OutputPaths { shell_gc_root } = &root_paths;
     if root_paths.all_exist() {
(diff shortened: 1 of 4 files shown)


curiousleo:
Relates to https://github.com/target/lorri/issues/287.

src/ops/info.rs
```diff
@@ -1,17 +1,20 @@
...
-
+pub fn main(project: Project) -> OpResult {
+    info!("lorri version: {}", crate::VERSION_BUILD_REV);
+    let root_paths = Roots::from_project(&project).paths();
+    let OutputPaths { shell_gc_root } = &root_paths;
+    if root_paths.all_exist() {
+        info!("gc roots exist"; "shell_gc_root" => ?shell_gc_root.as_os_str());
+    } else {
+        info!("gc roots do not exist"; "shell_gc_root" => ?shell_gc_root.as_os_str());
+    }
```
grahamc: I think since the entire purpose of the info command is to print this info, it should use `println` instead of a logging mechanism.

curiousleo: Fixed.

curiousleo:
PTAL

Profpatsch:
Looks good!

PR #288, opened 2020-01-17
lorri shell

curiousleo:
## Overview

Resurrects `lorri shell`. The idea is to make it as easy as possible for people with existing projects with `shell.nix` files to get started with lorri. The pitch is "just use `lorri shell` instead of `nix-shell`".

With the approach taken in this PR, `lorri shell` will always launch into bash, regardless of the shell from which `lorri shell` was called. In a follow-up PR, we'll lift this restriction.

## Checklist


Profpatsch:
> rising from its grave  - [**comparison here.**](https://github.com/target/lorri/compare/429b17921055893fb41e1725cec642add76f2f96..1d2395bce298b73acfae10749e756ae3cae6386c#diff-045010dc06c130b7f500845ce7a44bb1)

what is haaaappeninnggg

src/ops/direnv/envrc.bash
```diff
@@ -169,5 +169,9 @@ elif [ -f "$EVALUATION_ROOT" ]; then
     # shellcheck disable=SC1090
     . "$EVALUATION_ROOT"
 fi
+if [ -f "$EVALUATION_ROOT/alias-export" ]; then
+    # shellcheck disable=SC1090
+    . "$EVALUATION_ROOT/alias-export"
```
curiousleo: This works: aliases set in the `shellHook` get exported and sourced. Unfortunately, functions defined in the `shellHook` do not get exported. So e.g. the `ci` alias in this project gives an error under `lorri shell` as of now:

```console
$ lorri shell
$ ci
ci_check: command not found
```

grahamc: Looks great! functions are never exported with lorri, so that is expected behavior. If we want a `ci` or `ci_check` etc. function to come through, best to make it a script.

src/ops/shell.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::path::PathBuf;
+use std::process::Command;
+
+/// See the documentation for lorri::cli::Command::Shell for more
+/// details.
+pub fn main(project: Project) -> OpResult {
+    warn!(
+        "lorri shell is very simplistic and not suppported at the moment. \
+         Please use the other commands."
+    );
```
grahamc: let's drop this

src/ops/direnv/envrc.bash
grahamc: Wait no, let's ditch this. I don't think we should support aliases: `lorri shell` will ideally use whatever the user's $SHELL is and supporting things like aliases makes that much harder.

src/ops/shell.rs
curiousleo: Done

src/ops/direnv/envrc.bash
curiousleo: I've removed the alias export / import stuff.

> ideally use whatever the user's $SHELL is

Okay, right now it's using `bashInteractive` from lorri's runtime closure. Should it be using $SHELL instead? In that case we would need a variant of `envrc.bash` for every supported shell.

curiousleo: It looks like `nix-shell` also drops you into `bashInteractive` regardless of the original shell.

```console
$ sh
$ ps -p $$
  PID TTY          TIME CMD
 1612 pts/0    00:00:00 sh
$ nix-shell
$ ps -p $$
  PID TTY          TIME CMD
 1638 pts/0    00:00:00 bash
```

I get the same when the starting point is a `fish` shell.

curiousleo:
Closing in favour of https://github.com/target/lorri/pull/295.

PR #295, opened 2020-01-23
lorri shell

curiousleo:
## Overview

Resurrects `lorri shell`. The idea is to make it as easy as possible for people with existing projects with `shell.nix` files to get started with lorri. The pitch is "just use `lorri shell` instead of `nix-shell`".

This version of `lorri shell` does not cache build results and does not attempt to launch from the cache. That will be done in a follow-up PR.

## Checklist

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,6 +6,12 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 429;
+        changes = ''
+          - Re-introduced `lorri shell` command.
+        '';
+      }
+      {
         version = 309;
         changes = ''
           - The deprecated `lorri shell` command was removed.
--- a/release.nix
+++ b/release.nix
@@ -6,12 +6,6 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
-        version = 429;
-        changes = ''
-          - Re-introduced `lorri shell` command.
-        '';
-      }
-      {
         version = 309;
         changes = ''
           - The deprecated `lorri shell` command was removed.
--- a/src/ops/shell.rs
+++ b/src/ops/shell.rs
@@ -1,0 +1,87 @@
+//! Open up a project shell
+
+use crate::builder;
+use crate::builder::RunStatus;
+use crate::nix::CallOpts;
+use crate::ops::error::{ExitError, OpResult};
+use crate::project::{roots::Roots, Project};
+use crossbeam_channel as chan;
+use slog_scope::debug;
+use std::io;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::time::{Duration, Instant};
+use std::{env, fs, thread};
+
+/// See the documentation for lorri::cli::Command::Shell for more
+/// details.
+pub fn main(project: Project) -> OpResult {
+    let shell = env::var("SHELL").expect("lorri shell requires $SHELL to be set");
+    debug!("using shell path {}", shell);
+
+    let tempdir = tempfile::tempdir().expect("failed to create temporary directory");
+    let mut bash_cmd = bash_cmd(project, tempdir.path())?;
+    debug!("bash"; "command" => ?bash_cmd);
+    bash_cmd
+        .args(&["-c", &format!("exec {}", shell)])
+        .status()
+        .expect("failed to execute bash");
+    Ok(())
+}
+
+/// Instantiates a `Command` to start bash.
+pub fn bash_cmd(project: Project, tempdir: &Path) -> Result<Command, ExitError> {
+    let (tx, rx) = chan::unbounded();
+    thread::spawn(move || {
+        eprint!("lorri: building environment");
+        let mut last = Instant::now();
+        for msg in rx {
+            // Set the maximum rate of the "progress bar"
+            if last.elapsed() >= Duration::from_millis(500) {
+                eprint!(".");
+                io::stderr().flush().unwrap();
+                last = Instant::now();
+            }
+            debug!("build"; "message" => ?msg);
+        }
+        eprintln!(". done");
+    });
+
+    let run_result = builder::run(tx, &project.nix_file, &project.cas)
+        .map_err(|e| ExitError::temporary(format!("build failed: {:?}", e)))?;
+    let build = match run_result.status {
+        RunStatus::Complete(build) => Roots::from_project(&project)
+            .create_roots(build)
+            .map_err(|e| ExitError::temporary(format!("rooting the environment failed: {:?}", e))),
+        e => Err(ExitError::temporary(format!("build failed: {:?}", e))),
+    }?;
+
+    let init_file = tempdir.join("init");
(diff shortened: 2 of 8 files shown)


curiousleo:
Addressing https://github.com/target/lorri/pull/288#pullrequestreview-347344521 here:

> 1. change the PS1 to something else to tell the user it worked

- I have my PS1 configured just the way I want it, and don't like it when it gets overridden.
- PS1 works for bash and zsh, but the equivalent settings [in fish](https://stackoverflow.com/a/24582165/3507119) and [in elvish](https://stackoverflow.com/q/56228984/3507119) work very differently, so this would introduce work we'd have to repeat for each shell we want to support.

> 2. can we print the evaluation log data while it goes?

I've implemented something slightly different here: this will keep printing dots (at most one per 500ms) as build messages come in. It's a poor man's progress bar, but the point is that it shows the user that _something_ is happening. Let me know how it feels!

---

I like the ideas for `--cached`. I'll look at that once this is merged.

curiousleo:
PTAL

release.nix
```diff
@@ -5,6 +5,12 @@
   changelog = {
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
+      {
+        version = 429;
+        changes = ''
+          - Re-introduced `lorri shell` command.
```
Profpatsch: Would add a reasoning here so users see it on update.

src/ops/shell.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::time::{Duration, Instant};
+use std::{env, fs, thread};
+
+/// See the documentation for lorri::cli::Command::Shell for more
+/// details.
+pub fn main(project: Project) -> OpResult {
+    let shell = env::var("SHELL").expect("lorri shell requires $SHELL to be set");
```
Profpatsch: Can we handle non-POSIX `SHELL` like `fish`?

src/ops/shell.rs
```diff
@@ -0,0 +1,87 @@
...
+        ),
+    )
+    .expect("failed to write shell output");
+
+    debug!("building bash via runtime closure"; "closure" => crate::RUN_TIME_CLOSURE);
+    let bash_path = CallOpts::expression(&format!("(import {}).path", crate::RUN_TIME_CLOSURE))
+        .value::<PathBuf>()
+        .expect("failed to get runtime closure path");
+
+    let mut cmd = Command::new(bash_path.join("bash"));
```
Profpatsch: So we are using `bash -c` to manually get the stuff that would usually be loaded by the `direnv` environment, and then we call the user shell?

A few docstrings and inline strings would be good for explanation.

tests/shell/main.rs
```diff
@@ -0,0 +1,34 @@
...
+    assert_eq!(
+        // The string conversion means we get a nice assertion failure message in case stdout does
+        // not match what we expected.
+        String::from_utf8(output.stdout).expect("stdout not UTF-8 clean"),
+        "my_env_value\n"
+    );
+}
+
+fn project(name: &str, cache_dir: &Path) -> Project {
+    let test_root = PathBuf::from_iter(&[env!("CARGO_MANIFEST_DIR"), "tests", "shell", name]);
```
Profpatsch: Instead of that you can just use a `ContentAddressable` on a tempdir and reify an inline string from the test into a file. Then the files contents are directly here in the test and it’s easier to read.

grahamc:
```
[grahamc@Petunia:~/projects/github.com/grahamc/network]$ lorri shell
lorri: building environment........... done

[grahamc@Petunia:~/projects/github.com/grahamc/network]$
```

Looking better!


```
[grahamc@Petunia:~/projects/github.com/grahamc/network]$ lorri shell
lorri: building environment. done
Jan 23 21:25:35.219 ERRO build failed: FailedAtInstantiation

```

hm can we get some more specific error output?


---


I know you have your PS1 configured the way you like, but I think we simply can't not give any visual indicator. I have found myself closing terminals by mistake thinking I'm in a lorri shell already, and similarly being several lorri shells deep.

curiousleo:
@Profpatsch wrote:
> This is mostly the old code revived, right? Looks good in principle, a few comments.

That's how it started. At this point, the two versions have very much diverged: https://github.com/target/lorri/compare/429b17921055893fb41e1725cec642add76f2f96..curiousleo:lorri-any-shell-3#diff-045010dc06c130b7f500845ce7a44bb1

src/ops/shell.rs
curiousleo: I've tested this with `fish` and `elvish`. One issue is that sub-shells appear not to overwrite `$SHELL`: my default shell is bash, and when I create a fish sub-shell, `$SHELL` inside the fish shell still points to the bash binary. But my assumption is that when fish is your default shell, `$SHELL` points to the fish binary and will thus get picked up by `lorri shell`.

To test whether `lorri shell` works with fish and elvish in principle, I manually set `$SHELL` to their respective binary paths and then launched `lorri shell`, which successfully launched fish and elvish.

src/ops/shell.rs
curiousleo: Yup that's it. I'll add some documentation around here.

tests/shell/main.rs
curiousleo: I'll give that a go, thanks.

curiousleo:
[@grahamc wrote:](https://github.com/target/lorri/pull/295#issuecomment-577966057)
> I know you have your PS1 configured the way you like, but I think we simply can't not give any visual indicator. I have found myself closing terminals by mistake thinking I'm in a lorri shell already, and similarly being several lorri shells deep.

[@curiousleo wrote:](https://github.com/target/lorri/pull/295#issuecomment-577777753)
> PS1 works for bash and zsh, but the equivalent settings in fish and in elvish work very differently, so this would introduce work we'd have to repeat for each shell we want to support (something we've managed to avoid so far using the exec $SHELL trick).

E.g. here is how `venv` does this [for bash/zsh](https://github.com/python/cpython/blob/65ecc390c1fa5acdd6348ae3f9843bbdcd8870d1/Lib/venv/scripts/common/activate#L55-L59), [for fish](https://github.com/python/cpython/blob/65ecc390c1fa5acdd6348ae3f9843bbdcd8870d1/Lib/venv/scripts/posix/activate.fish#L50-L61) and [for csh](https://github.com/python/cpython/blob/65ecc390c1fa5acdd6348ae3f9843bbdcd8870d1/Lib/venv/scripts/posix/activate.csh#L19-L21).

src/ops/shell.rs
curiousleo: Feel free to try this yourself with fish! :)

release.nix
grahamc: Let's delete this entry until #297 is done.

curiousleo: Done

grahamc:
LGTM, I moved further steps in to #297 :)

Issue #311, opened 2020-01-30
Specify semantic versioning policy for lorri


curiousleo:
After the [1.0[.0] release](https://github.com/target/lorri/issues/306), at some point we'll want to make another release. At that point we'll ask ourselves: what version should it have - that is, should we increment the patch, minor, or major version?

Let's specify a versioning policy for lorri that answers this question based on the changes since the last release.

References:
- https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field
- https://semver.org/
- https://github.com/target/lorri/issues/306#issuecomment-579955000
- https://github.com/target/lorri/issues/306#issuecomment-580130915

PR #312, opened 2020-01-30
Specify versioning scheme

curiousleo:
Closes https://github.com/target/lorri/issues/311.

## [Preview](https://github.com/curiousleo/lorri/blob/version-policy/MAINTAINERS.md#versioning-scheme)

See https://github.com/target/lorri/issues/311 for context.

## Checklist

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -10,28 +10,36 @@ In order to apply this scheme to lorri, we need to define what "incompatible
 change" means for lorri, a command line tool.
 
 Let's clarify some terminology. lorri is a command line tool with multiple
-**subcommands**. Some of them are **internal**. There are mandatory and
-optional **command line options**, which can be attached to the top-level
+**subcommands**. Those subcommands are either _external_ or _internal_.
+Internal subcommands' names end in an underscore. There are _mandatory_ and
+_optional_ **command line options**, which can be attached to the top-level
 command or to subcommands.
 
 Subcommands fall into two categories based on the intended consumer of their
-outputs: their **outputs** may be for human consumption or machine-readable.
+outputs: their **outputs** may be for _human consumption_ or
+_machine-readable_.
 
-**When to release a new major version**
+### When to make a major release
 
 Increment the `MAJOR` version if since the last released version,
-- a non-internal subcommand was removed or renamed, or
-- a top-level command line option was removed or renamed, or
-- a previously optional top-level command line option was made mandatory, or
-- a command line option of a non-internal subcommand was removed or renamed, or
-- a previously optional command line option of a non-internal subcommand was
-  made mandatory, or
-- the output of a non-internal machine-readable subcommand changed, or
+- for any _top-level_ command line option:
+  - the command line option was removed or renamed, or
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -1,5 +1,42 @@
...
+- a non-internal subcommand was removed or renamed, or
+- a top-level command line option was removed or renamed, or
+- a previously optional top-level command line option was made mandatory, or
+- a command line option of a non-internal subcommand was removed or renamed, or
+- a previously optional command line option of a non-internal subcommand was
+  made mandatory, or
+- the output of a non-internal machine-readable subcommand changed, or
+- it is conceivable that a project that could previously be built successfully
+  now fails to build with lorri, unless the previous behaviour is considered a
+  bug.
```
grahamc: This goes through some gymnastics to basically say non-internal commands aren't APIs and maybe you're going to poke your eye out on them. Maybe make the statement about how the API of an internal command should be treated at the top, and then say "for all public commands..."

MAINTAINERS.md
```diff
@@ -1,5 +1,42 @@
...
+
+In any other case, increment the `MINOR` version.
+
+Since lorri is exclusively built with Nix and its runtime dependencies are
+captured in its runtime closure, changing a build-time or runtime dependency
+does not in itself necessitate a major release.
```
grahamc: :+1:

Profpatsch:
We can just stick with semver, it should cleanly apply to CLI projects:

* Internal change or fix -> patch
* New feature without incompatible change -> minor
* Breaking change -> major

MAINTAINERS.md
```diff
@@ -1,5 +1,55 @@
 # Maintaining lorri
 
+## Versioning scheme
+
+The versioning scheme for lorri is `MAJOR.MINOR`, where `MAJOR` and `MINOR` are
+non-negative integers without leading zeroes.
```
grahamc:
```suggestion
non-negative integers without leading zeroes. We follow the same rules as semver,
and remainder of this section describes exactly what that means for lorri.
```

curiousleo: I mentioned semver in a previous iteration. Then I took it out, because I think that it doesn't really help.
- semver explicitly uses `MAJOR.MINOR.PATCH` whereas this document proposes `MAJOR.MINOR`
- semver talks about "API backwards compatibility", which only applies in an abstract sense to a command line tool

For those reasons, I came to the conclusion that "it's like semver, but ..." is less clear than just explaining the scheme directly. It's definitely semver-inspired, but I think of semver as a complete versioning scheme that one should adopt or not adopt. "Like semver, but ..." just doesn't really aid understanding in my opinion.

grahamc: Seems fine enough to me. We're not in the library business or production operations business. Maybe one day this will come back to bite us, but if it does, we can do something different then.

MAINTAINERS.md
curiousleo: Yep, that wording was really awkward. Fixed, thanks!

PR #314, opened 2020-01-30
How to publish lorri on nixpkgs

curiousleo:
Relates to https://github.com/target/lorri/issues/269.

## [Preview](https://github.com/curiousleo/lorri/blob/how-to-release-on-nixpkgs/MAINTAINERS.md#publishing-a-release-on-nixpkgs)

Documents how to publish lorri on nixpkgs.

## Checklist

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -4,6 +4,26 @@
 
 TODO: https://github.com/target/lorri/issues/269
 
+## Publishing a release on [nixpkgs][]
+
+Currently (2020-01-30), lorri is available in the `nixos-unstable` and
+`nixos-19.09` release channels, which correspond to the `master` and
+`release-19.09` branches in the [nixpkgs][] repository, respectively.
+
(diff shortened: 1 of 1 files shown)


grahamc:
I think it would be good to say that in the future, after 1.0, it is our goal to keep Lorri building with tools available in the current latest stable release. (ie: 19.09, or 20.03 once that is stable.)

curiousleo:
@grahamc wrote:
> I think it would be good to say that in the future, after 1.0, it is our goal to keep Lorri building with tools available in the current latest stable release. (ie: 19.09, or 20.03 once that is stable.)

In that case, we should probably pin nixpkgs to the latest stable release to make sure we're compatible: https://github.com/target/lorri/blob/f7bd9b0029d173f8eba776e0753aae1684fe247f/nix/update-dependencies.sh#L6-L7

This should then use `refs/heads/nixos-<earliest supported stable channel>` instead of `refs/heads/nixos-unstable`.

PR #320, opened 2020-02-01
Event Streaming (take 2)

### LINKED ISSUE #145: Indicate that Lorri is done building for a project

nyarly:
Closes #145 #78 #79.

This PR should replace #183 - if this is acceptable, please close that PR.

## Overview

Adds a `lorri internal__stream_events` subcommand that returns a continuous stream json formatted representations of events. Each event is its own \n separated line. The command takes an option to toggle current state reporting vs. live events (or both), which is useful for the motivating use cases.

## Checklist


curiousleo:
I left a little review (https://github.com/target/lorri/pull/183#pullrequestreview-352067814) and a couple of comments (https://github.com/target/lorri/pull/183#issuecomment-581301025, https://github.com/target/lorri/pull/183#issuecomment-581325826) on #183 before seeing this new PR. I was going to suggest you rebase on top of `master` at some point before the PR gets merged, and that's what you seem to have done - great! :)

Just as a note: I have no issue at all with you force-pushing to the branch of an existing PR like #183, IMO there's no need from the reviewer's point of view to create a new PR. But I'm happy to follow along.

curiousleo:
I had another think about this and spoke to @grahamc. Here's what I suggest we do:

1. We merge this PR (#320), as it is right now. It adds a useful feature and people can start experimenting. It's also explicitly labelled as "internal" so people don't expect the output format to be set in stone.
2. We address https://github.com/target/lorri/pull/183#issuecomment-581325826 by separating out the representation of the output format from internal structs and enums.
3. We address the two remaining items listed here: https://github.com/target/lorri/pull/183#issuecomment-581301025

I'm happy to help as much or as little as you like with each step.

nyarly:
> I had another think about this and spoke to @grahamc. Here's what I suggest we do:
...
> @nyarly how does that sound to you?

@curiousleo That plan sounds great to me. I would really like to get a nicer version of the example scripts into `contrib`. The comments regarding an explicit interface make complete sense, and I'm happy to make inroads there as well. Both changes strike me as nicely scoped and straightforward.

curiousleo:
> > I had another think about this and spoke to @grahamc. Here's what I suggest we do:
> > ...
> > @nyarly how does that sound to you?
> 
> @curiousleo That plan sounds great to me. I would really like to get a nicer version of the example scripts into `contrib`. The comments regarding an explicit interface make complete sense, and I'm happy to make inroads there as well. Both changes strike me as nicely scoped and straightforward.

Excellent - congratulations, and thank you for your contributions so far!

I've created https://github.com/target/lorri/issues/324 to keep track of the follow-up items we've discussed.

symphorien:
Is this command supposed to return constant output ?
I have a shell.nix failing to evaluate, and running `lorri internal__stream_events --kind snapshot` repeatedly sometimes includes a like about the failed evaluation, and sometimes does not.

nyarly:
`--kind snapshot` reports the current understanding of the server of your builds - it should change over time as your builds do.

symphorien:
I mean:
I have 1 shell.nix failing to evaluate, and 3 shells already built.
If I continuously run `lorri internal__stream_events --kind snapshot` **without changing anything on the file system**, the failing shell is only reported about half the time. That does not look right to me.

Issue #306, opened 2020-01-29
Release 1.0


grahamc:

curiousleo:
Related: https://github.com/target/lorri/issues/269

curiousleo:
Does "1.0" mean we're ditching the "number of commits" version scheme that's currently built into `release.nix` and `lorri upgrade`?
Also see https://github.com/target/lorri/issues/269#issuecomment-579822694

grahamc:
I think we should keep those numbers, and also add a real version number. Is that silly?

curiousleo:
> I think we should keep those numbers, and also add a real version number. Is that silly?

I think that has the potential to confuse users. My radical proposal for this would be:
- remove `lorri upgrade` and changelog parsing; I don't think we need it
- take `release.nix` and turn it into a `CHANGELOG.md` with entries for the so-far released versions `unstable-2019-10-30` and `unstable-2020-01-09`
- from this point on, update `CHANGELOG.md` like a conventional changelog, [e.g. this one](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md).

Profpatsch:
>     * remove `lorri upgrade` and changelog parsing; I don't think we need it

I use it regularly to update lorri to current master, it’s great (also for trying out in-progress branches once https://github.com/target/lorri/pull/168 is merged)

Profpatsch:
Apart from that we should make an effort to check that every part of the command line is as stable as possible, and what parts of the command line are intended for machine-readable output; e.g. we wouldn’t want people to start parsing the output of `lorri info` or even `lorri watch`.

curiousleo:
> I use it regularly to update lorri to current master, it’s great (also for trying out in-progress branches once #168 is merged)

I think we've discussed this before - you use `lorri self-upgrade`, I don't (`git checkout <branch> && nix-build .` works well enough for me), therefore I see it mainly as something that increases the cost of maintaining lorri whereas you see it as something that adds value. I don't see a way towards a consensus on this.

curiousleo:
> Apart from that we should make an effort to check that every part of the command line is as stable as possible, and what parts of the command line are intended for machine-readable output; e.g. we wouldn’t want people to start parsing the output of `lorri info` or even `lorri watch`.

Yes. It's a bit unfortunate that this discussion is happening on two issues at once -- also see https://github.com/target/lorri/issues/269#issuecomment-579822694:

[...]

PR #323, opened 2020-02-04
Prepare 1.0 release

curiousleo:
Part of https://github.com/target/lorri/issues/306.

## Overview

Bumps the version to 1.0. `lorri info` shows this version now, rather than the revision count. Updates nixpkgs also for good measure.

```console
$ lorri info --shell-file shell.nix 
lorri version: 1.0
[...]
```

## Checklist


build.rs
```diff
@@ -5,38 +5,44 @@ use std::fs::File;
...
+
 /// Number of revisions in the Git tree.
 pub const VERSION_BUILD_REV: usize = {};
 
 /// Run-time closure parameters. This argument points to a file
 /// generated by ./nix/runtime.nix in Lorri's source.
 pub const RUN_TIME_CLOSURE: &str = "{}";
 "#,
-            rev_count, run_time_closure
+            version, rev_count, run_time_closure
```
grahamc: With the size of this block, I'm thinking it might be good to use named parameters:

```
format!("{argument}", argument = "test");   // => "test"
```

What do you think about that?

curiousleo: TIL :) Thanks!

curiousleo:
> lgtm, what do you think about the comment?

I just learned that `format!` accepts named arguments!

PR #340, opened 2020-03-02
Ci use cachix on macos

Profpatsch:
let’s see whether we can get cachix on macOS running again.

cc @curiousleo @domenkozar


Profpatsch:
It seems to work and ran something like 40 minutes on macOS.

Restarting to check whether we can actually fetch the packages that it build from cachix.

domenkozar:
Why was there no branch build triggered?

Profpatsch:
8 minutes. Sweet.

> Why was there no branch build triggered?

Because we don’t have them enabled.

PR #342, opened 2020-03-03
Ci build on master

Profpatsch:
Fix a few CI linting errors and enable master builds.

I’m not 100% sure whether this disables PR builds, let’s see …

PR #344, opened 2020-03-03
CI: make job names different

Profpatsch:
This is more of a dummy PR to try out whether master builds after https://github.com/target/lorri/pull/342


Profpatsch:
@grahamc I had to fix the bug110 test, seems to work.

Issue #341, opened 2020-03-03
build queue is not debounced anymore


Profpatsch:
**Describe the bug**

When an event triggers a build, previously what happened was:

- If there’s no build running, the build is started
- If a build is already running, the build is queued *iff there’s no other build queued for that file yet*

**To Reproduce**

Change a file multiple times and watch every change be translated in a (superficial) evaluation

**Expected behavior**

Debounce the builds

$ lorri info
current master
```

**Comment by michaelpj** [comment]
This is especially painful if your evaluation does a bunch of IFD! :skull:

**Comment by nomeata** [comment]
I just changed branch on one of my lorri projects, and i tlooks like I see a 
```
Apr 29 09:52:55 kirk lorri[20859]: Apr 29 09:52:55.362 INFO build status, message: BuildEvent(Started { nix_file: Shell("/home/jojo/project/shell.nix"), reason: FilesChanged(["/home/jojo/dfinity/project/sources.json"]) })
```
followed by `BuildResults` event for every single `nix` file in the project. This is probably the present bug, is it?

curiousleo:
@Profpatsch, could you clarify this bug report?

- Is the "previous" behaviour the expected behaviour? If so, let's call it that.
- In the observed behaviour, do you see _simultaneous_ evaluations? Or do they pile up in a queue and then get evaluated one after another? This would help clarify the relationship between this issue and https://github.com/target/lorri/issues/389.

Profpatsch:
>     * Or do they pile up in a queue and then get evaluated one after another?

^ this one

I did some preliminary refactoring a while ago, but didn’t come around to do the actual fix.

Nadrieril:
I _think_ this PR is an attempt at fixing this: https://github.com/target/lorri/pull/465

PR #345, opened 2020-03-06
Refactor build loop a bit

Profpatsch:
This is the first step to solving https://github.com/target/lorri/issues/341, which is refactoring the build loop. See the commits, there should be no changes in functionality in this code.

cc @curiousleo
with help from @NinjaTrappeur

### DIFF
--- a/src/build_loop.rs
+++ b/src/build_loop.rs
@@ -78,7 +78,6 @@ impl<'a> BuildLoop<'a> {
     /// still running, it is finished first before starting a new build.
     #[allow(clippy::drop_copy, clippy::zero_ptr)] // triggered by `select!`
     pub fn forever(&mut self, tx: chan::Sender<LoopHandlerEvent>, rx_ping: chan::Receiver<()>) {
-        let send = |msg| tx.send(msg).expect("Failed to send an event");
         let translate_reason = |rsn| match rsn {
             Ok(rsn) => rsn,
             // we should continue and just cite an unknown reason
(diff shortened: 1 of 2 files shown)


src/build_loop.rs
```diff
@@ -95,66 +94,91 @@ impl<'a> BuildLoop<'a> {
             }
         };
 
+        let mut output_paths;
         // The project has just been added, so run the builder in the first iteration
-        let mut reason = Some(Event::Started {
+        let reason = Event::Started {
```
curiousleo: I would inline `reason` (used once) to remove one variable from the scope.

src/build_loop.rs
```diff
@@ -95,66 +94,91 @@ impl<'a> BuildLoop<'a> {
...
             nix_file: self.project.nix_file.clone(),
             reason: Reason::ProjectAdded(self.project.nix_file.clone()),
-        });
-        let mut output_paths = None;
+        };
+        output_paths = self.once_with_send(&tx, reason);
```
curiousleo:
```suggestion
        let mut output_paths = self.once_with_send(&tx, reason);
```

src/build_loop.rs
```diff
@@ -95,66 +94,91 @@ impl<'a> BuildLoop<'a> {
             }
         };
 
+        let mut output_paths;
```
curiousleo: No need for this separate declaration.

curiousleo:
>  Fix build loop channel debounce
Would you mind changing the title of this PR? It's out of sync with the description, which makes it clear that this is a preparatory refactor that does not actually change behaviour.

Profpatsch:
> Would you mind changing the title of this PR? It's out of sync with the description, which makes it clear that this is a preparatory refactor that does not actually change behaviour.

yeah, that’s just github expanding the branch name :) Changed

Issue #324, opened 2020-02-04
Event stream: follow-up


curiousleo:
This issue records follow-up work for the "event stream" feature, based on https://github.com/target/lorri/pull/320#issuecomment-581421805.

- [x] Merge event stream PR
   PR: https://github.com/target/lorri/pull/320
- [ ] Separate out representation of event stream output format
   Context: https://github.com/target/lorri/pull/183#issuecomment-581325826
- [ ] Document event stream uses with examples
   Context: https://github.com/target/lorri/pull/183#discussion_r373391813, https://github.com/target/lorri/pull/183#issuecomment-581301025
- [ ] Strengthen typed representation of internal event stream messages
   Context: https://github.com/target/lorri/pull/183#discussion_r373508954, https://github.com/target/lorri/pull/183#issuecomment-581301025

This work does not block the 1.0 release (https://github.com/target/lorri/issues/306). The follow-up items don't need to be tackled in any particular order. Separating out the representation of the event stream output format is necessary before the event stream command can be transitioned to non-internal (if we wish to do that).

PR #346, opened 2020-03-08
Conversions for JSON DTO

nyarly:
First task of issue #324

## Overview

Creates a dedicated DTO format for JSON, so we can use that format as part of the public API.

## Checklist

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,6 +6,12 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 511;
+        changes = ''
+          Stabilized the JSON format for `internal__stream_events`
+        '';
+      }
+      {
         version = 476;
         changes = ''
           Introduces internal subcommand `lorri internal__stream_events`
--- a/src/ops/stream_events.rs
+++ b/src/ops/stream_events.rs
@@ -1,5 +1,4 @@
 //! Run to output a stream of build events in a machine-parseable form.
-use crate::build_loop::Event;
 use crate::ops::{
     error::{ok, ExitError, OpResult},
     get_paths,
@@ -11,6 +10,9 @@ use std::convert::TryInto;
 use std::str::FromStr;
 use std::thread;
 use std::time::Duration;
+
+mod json_representation;
+use json_representation::Event;
 
 /// Options for the kinds of events to report
 #[derive(Debug)]
--- a/src/ops/stream_events/json_representation.rs
+++ b/src/ops/stream_events/json_representation.rs
@@ -1,0 +1,169 @@
+use std::convert::{TryFrom, TryInto};
+
+use crate::rpc;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub(super) enum Event {
+    SectionEnd,
+    /// A build has started
+    BuildStarted {
+        /// The shell.nix file for the building project
+        shell_file: String,
+        /// The reason the build started
+        reason: Reason,
+    },
+    /// A build completed successfully
+    BuildCompleted {
+        /// The shell.nix file for the building project
+        shell_file: String,
+        /// The result of the build
+        result: Vec<String>,
+    },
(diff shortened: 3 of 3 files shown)


curiousleo:
Hi @nyarly, thanks for getting back to this!

In https://github.com/target/lorri/pull/183#issuecomment-581325826, I wrote
> We can use Varlink type definitions to describe the output format. Describing a JSON schema is something Varlink is pretty good at, in my opinion. People who want to build tooling for lorri using event_stream may find it easier to read the Varlink schema than Rust structs and enums. Finally, changes to this definition would be very obvious, this would pretty much rule out making output format changes by accident.

You went for Rust structs instead of Varlink definitions for the public API definition. I'd be curious about your perspective on the pros and cons of a Varlink vs. Rust representation of the public API.

nyarly:
@curiousleo I think I missed that comment at the time. Let me look into using Varlink for this.

nyarly:
One preliminary concern: is there a sensible way to make clear that there's the JSON Varlink (part of the public interface) and the internal Varlink (which is private and changeable)?

curiousleo:
> One preliminary concern: is there a sensible way to make clear that there's the JSON Varlink (part of the public interface) and the internal Varlink (which is private and changeable)?

There should be a new, separate Varlink file for the public interface. The existing Varlink file with the private interface should probably get a big banner at the top saying "this is a private interface; don't use it; may change at any time without notice".

nyarly:
So, two varlink interfaces. New one called `com.target.lorri.public` (or `.eventstream` ?)

I'm looking today at how varlink.rs provides serialization without wire protocol...

nyarly:
...oh, the generated message structs are `#[derive(Serialize)]` Okay, this seems like it'll be straightforward to implement. The varlink messages just replace the handwritten structs, with the `TryInto` implementations applying to the message structs instead.

nyarly:
Okay, so here's a line of thought @curiousleo @Profpatsch - I started working through a varlink Events message for serialization to JSON. It'll be straightforward to replace the Rust structs with that message, I think. 

But, wouldn't it make sense to fully pull the Event interface out of the internal varlink definition and have the daemon use the same messaging? If we're exposing the message via varlink, it would be surprising (I would be surprised, at least) not to make direct varlink integration possible, and for that you need a method and a place to make the requests.

I think I'd be happy with `stream_events` being a very thin layer that just outputs the messages to stdout for pipeline construction since that was always the design. Having it simply reserialize the messages it receives would further simplify the idea.

But, it would fix the Event message in the interface... I'm not sure that's terrible, since it's effectively fixed already. No significant change can be useful without exposing it in the output, anyway.

In this case, I'd also argue against changing the structure of the message, since the "type safety" argument isn't as strong when the message is meant to be consumed by jq parsers.

curiousleo:
> But, wouldn't it make sense to fully pull the Event interface out of the internal varlink definition and have the daemon use the same messaging? If we're exposing the message via varlink, it would be surprising (I would be surprised, at least) not to make direct varlink integration possible, and for that you need a method and a place to make the requests.

I am not 100% sure what you're proposing concretely here. What I want to make sure is that (1) the _public_ interface is well-defined, ideally as Varlink in a separate file; (2) no internal declarations (Varlink or Rust) leak into the public interface.

> In this case, I'd also argue against changing the structure of the message, since the "type safety" argument isn't as strong when the message is meant to be consumed by jq parsers.

I don't think this argument is very strong. Every JSON interface has a "type" of some sort - when you call a JSON API, there are usually fields that you expect in the answer. The simpler the answer to the question "which fields can I expect?", the better. The proposed structure in https://github.com/target/lorri/pull/183#discussion_r373508954 makes this answer very simple, which is useful no matter where the output is used.

Profpatsch:
As far as I can see, all of this is included in the current lorri release.

We haven’t made the json format stable yet, but we could in the future.

Profpatsch:
For reference: current lorri release: https://github.com/nix-community/lorri/releases/tag/1.4.0

PR #348, opened 2020-03-14
ci: use Github actions

lovesegfault:
This is a best-effort attempt at porting the CI to use GitHub actions. I expect this to be incomplete still and am looking for feedback to improve it.

cc. @grahamc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,78 @@
+name: CI
+on:
+  push:
+    branches: [ "**" ]
+  pull_request:
+    branches: [ "**" ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Nix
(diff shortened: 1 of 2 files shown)


curiousleo:
Hi @lovesegfault, thanks for your contribution! Did you discuss this change with @grahamc?

My first take:
- `-350 +79` is a great change: the current build setup is clearly more complex than it needs to be.
- Building CI with nix using a pinned nixpkgs ([see here](https://github.com/target/lorri/blob/master/nix/update-dependencies.sh#L7-L9)) has the big advantage that we know lorri will build for that channel (nixos-19.09 currently). For example, we know that lorri will build against the Rust version shipped in that nixpkgs channel. Nixpkgs is the main way lorri is distributed right now, so this matters. In the GitHub CI action, it is not clear to me which Rust version is actually used.

Profpatsch:
I don’t know whether we can enable Github actions here in the first place, since `target` probably has strict rules on what CI is allowed for its repositories.

Profpatsch:
Can you resurrect the travis files? I’d like to have both in parallel to see if github actions bring an improvement. Also, like mentioned above, to see whether github actions are enabled in the target orga in the first place.

Profpatsch:
Gonna merge and see whether it does something.

PR #350, opened 2020-03-14
make nixpkgs overridable in the shell.nix template

veprbl:
## Overview

The default template is missing one useful feature that would allow users to evaluate their shells with modified packages set. The approach to fixing this is to update the default template to take `pkgs` attrset via a function argument.

## Checklist

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,6 +6,14 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 510;
+        changes = ''
+          - The shell.nix template used by `lorri init` was changed to take
+            `pkgs` as an argument with import of `<nixpkgs>` used as the
+            default value.
+        '';
+      }
+      {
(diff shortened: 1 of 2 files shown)


veprbl:
This appears to be a very minor change. Let me know if I should mention it in the release.nix

curiousleo:
Thanks for your contribution! This seems like a great change: backwards compatible and giving users more options. Great!

> This appears to be a very minor change. Let me know if I should mention it in the release.nix

I think it would be good to mention it: this is user facing, so users should know about it.

veprbl:
@curiousleo I've updated the PR. For the version I used result of `git log --pretty=%h | wc -l` for the master branch, let me know if this is incorrect.

curiousleo:
> For the version I used result of `git log --pretty=%h | wc -l` for the master branch, let me know if this is incorrect.

That's exactly right. Thanks!

PR #353, opened 2020-03-20
fix(cli): display internal commands in a subcommand #343

### LINKED ISSUE #343: Add an option to list internal commands
With the change to `internal__`, the plumbing commands are completely unaccessible from the `help` output. Which is sad, because then users cannot use them to experiment with unstable features.

Let’s add an option to list the internal commands and make it clear from the toplevel helptext that some commands are hidden by default.

bnjjj:
## Overview

Put each internal commands in a subcommands visible for all users.
close #343 

BTW should I remove the ping command ? (related to the TODO comment in file)

## Checklist

- [X] Updated the documentation (code documentation, command help, ...)
- [X] Amended the changelog in `release.nix` (see `release.nix` for instructions)

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -8,8 +8,7 @@ non-negative integers without leading zeroes.
 ### Terminology
 
 lorri is a command line tool with multiple **subcommands**. Those subcommands
-are either _external_ or _internal_.  Internal subcommands' names start with
-"internal\_\_". There are _mandatory_ and _optional_ **command line options**,
+are either _external_ or _internal_. There are _mandatory_ and _optional_ **command line options**,
 which can be attached to the top-level command or to subcommands.
 
 Subcommands fall into two categories based on the intended consumer of their
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -8,8 +8,8 @@ non-negative integers without leading zeroes.
 ### Terminology
 
 lorri is a command line tool with multiple **subcommands**. Those subcommands
-are either _external_ or _internal_. There are _mandatory_ and _optional_ **command line options**,
-which can be attached to the top-level command or to subcommands.
+are either _external_ or _internal_. _internal_ subcommands are now all grouped in `lorri internal`.
+There are _mandatory_ and _optional_ **command line options**, which can be attached to the top-level command or to subcommands.
 
 Subcommands fall into two categories based on the intended consumer of their
 outputs: their **outputs** may be for _human consumption_ or
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,9 +6,7 @@
 // bit more typing, but the long name makes the intent much more obvious. The only short option
 // right now is `-v` for verbosity, and it should probably stay that way.
 //
-// Prefix internal commands with `internal__` and hide them from the auto-generated help using
-// `raw(setting = "structopt::clap::AppSettings::Hidden")`. See MAINTAINERS.md for details on
-// internal and non-internal commands.
+// See MAINTAINERS.md for details on internal and non-internal commands.
 
 use std::path::PathBuf;
 
@@ -43,13 +41,6 @@ pub enum Command {
     #[structopt(name = "shell")]
     Shell(ShellOptions),
 
-    /// (internal) Used internally by `lorri shell`
-    #[structopt(
-        name = "internal__start_user_shell",
-        raw(setting = "structopt::clap::AppSettings::Hidden")
-    )]
-    StartUserShell_(StartUserShellOptions_),
-
     /// Build project whenever an input file changes
     #[structopt(name = "watch")]
     Watch(WatchOptions),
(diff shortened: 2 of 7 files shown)


src/cli.rs
```diff
@@ -133,6 +120,22 @@ pub struct WatchOptions {
...
 }
 
+/// Sub-commands which lorri can execute for internal features
+#[derive(StructOpt, Debug)]
+pub enum Internal_ {
+    /// (internal) Used internally by `lorri shell`
+    #[structopt(name = "start_user_shell")]
+    StartUserShell_(StartUserShellOptions_),
+
+    /// (internal) Tell the lorri daemon to care about the current directory's project
```
Profpatsch:
```suggestion
    /// (plumbing) Tell the lorri daemon to care about the current directory's project
```

src/main.rs
```diff
@@ -112,21 +109,28 @@ fn run_command(log: slog::Logger, opts: Arguments) -> OpResult {
...
-        }
-
         Command::Init => {
             let _guard = without_project();
             init::main(TRIVIAL_SHELL_SRC, DEFAULT_ENVRC)
         }
+
+        Command::Internal { command } => {
+            match command {
+                // TODO: remove
```
Profpatsch: we can remove that TODO, I think ping is helpful for scripting.

Profpatsch:
cc @curiousleo

bnjjj:
@curiousleo My bad ! Thanks ! It's up to date now :)

MAINTAINERS.md
```diff
@@ -8,8 +8,7 @@ non-negative integers without leading zeroes.
 ### Terminology
 
 lorri is a command line tool with multiple **subcommands**. Those subcommands
-are either _external_ or _internal_.  Internal subcommands' names start with
-"internal\_\_". There are _mandatory_ and _optional_ **command line options**,
+are either _external_ or _internal_. There are _mandatory_ and _optional_ **command line options**,
```
curiousleo: I think it's worth mentioning how a user can tell: _internal_ subcommands are now all grouped in `lorri internal`.

src/ops/shell.rs
```diff
@@ -56,7 +56,7 @@ pub fn main(project: Project, opts: ShellOptions) -> OpResult {
     bash_cmd
         .args(&[
             "-c",
-            "exec \"$1\" internal__start_user_shell --shell-path=\"$2\" --shell-file=\"$3\"",
+            "exec \"$1\" internal start_user_shell --shell-path=\"$2\" --shell-file=\"$3\"",
```
curiousleo: Before this change in this file, the PR should have failed CI. So either
- there is a test that should have failed, but didn't - we need to investigate and fix (possibly in this PR)
- there is no test that covers this - we should add one (not in this PR)

bnjjj: Ok I checked about tests. First it seems the execution of this command wasn't really checked. I added some check about the command's state just here https://github.com/target/lorri/pull/353/files#diff-9cbb4bd47e53163015bb66adcbc9db2cR24

Then I checked if a test existed and covered these lines of code. But it didn't, in fact the test never covered this function. So I took the decision to execute a test with `assert_cmd` to launch it as a real user in his prompt in order to cover these lines of code.

Let me know what do you think about that ?

src/ops/shell.rs
```diff
@@ -68,7 +67,15 @@ pub fn main(project: Project, opts: ShellOptions) -> OpResult {
         ])
         .status()
         .expect("failed to execute bash");
-    Ok(())
+
+    if !status.success() {
+        Err(ExitError::panic(format!(
+            "cannot run lorri shell: fail to execute internal shell command (error: {})",
```
curiousleo:
```suggestion
            "cannot run lorri shell: failed to execute internal shell command (error: {})",
```

tests/shell/main.rs
```diff
@@ -1,18 +1,40 @@
+use assert_cmd::prelude::*;
```
curiousleo: The rest of the codebase uses explicit `use` statements without wildcards. IMO it is worth having explicit `use` statements where possible to give people reading the code a better idea of what's in scope. Is it reasonable to turn this `use` statement into explicit ones?

bnjjj: Ok updated :)

bnjjj:
> Very nice, thanks for increasing the test coverage! Did you check that this test would have caught the wrong command in `shell.rs`?
> 
> https://github.com/target/lorri/blob/c38e0f09c36a1b27d86eb01ba5672fdffa950641/src/ops/shell.rs#L59

In fact the previous test didn't catch it. Now the new one catch it :)

tests/shell/main.rs
```diff
@@ -1,18 +1,40 @@
...
+            "shell",
+            "--shell-file",
+            PathBuf::from(&project.nix_file)
+                .as_os_str()
+                .to_str()
+                .unwrap(),
+        ])
+        .current_dir(&tempdir)
+        .assert()
+        .success();
```
Profpatsch: I’m not convinced that we need to include 8 new transitive dependencies for this one line. Let’s just do it manually.

bnjjj: As it's in the dev-dependencies, is it really an issue ? It's only included for the test binary and not in the release binary. And I plan to add more tests about CLI so use this at different places.

Profpatsch: The issue here is that every transitive dependency blows up CI time, and we run these tests on CI. Every dependency is a potential point of problems in the future, and has to carry its (transitive) weight.

> And I plan to add more tests about CLI so use this at different places.

Is it an undue amount of code to just use `Command` and pattern-match on the result like we do in the other tests?

bnjjj: Ok 🙂 I updated the test

bnjjj:
up

Profpatsch:
Squashed & rebased on master. Merging once green.

PR #355, opened 2020-03-21
Don't watch /nix/store files (they can't change)

hamishmack:
## Overview

They are currently excluded in `add_path_recursively`, but not in
`add_path`.  This seems like an oversight.

I tested this change by switching the `debug!` calls in `add_path` to `info!` and observing that a large number of files in `/nix/store` were logged.

To fix this I have added the same guard used in `add_path_recursively` to `add_path`.

## Checklist

### DIFF
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -151,6 +151,10 @@ impl Watch {
     }
 
     fn add_path(&mut self, path: &PathBuf) -> Result<(), notify::Error> {
+        if path.canonicalize()?.starts_with(Path::new("/nix/store")) {
+            return Ok(());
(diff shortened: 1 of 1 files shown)


Profpatsch:
I’d like to improve the logic here, by having a `walk_dir` function which returns a `Vec<PathBuf>` with all recursive paths and then just iterating over that `Vec`, removing items we are not interested in.

Profpatsch:
@hamishmack take a look at https://github.com/target/lorri/pull/368, this should include your fix here.

hamishmack:
Cool.

PR #356, opened 2020-03-21
readDir: do not watch directories recursively

### LINKED ISSUE #152: Improve readDir watches

hamishmack:
Fixes #152 without breaking "copied source" (see #232).  To do this it:

  * Adds a `ReferencedPath` type to differentiate `Recusive` and
    `NotRecursive` path references found in the nix-instanciate output.
  * Applies the `reduce_paths` function only to the `Recursive`
    paths.
  * Uses the result of `reduce_paths` filter out any `NotRecursive`
    paths that would already be watched.
  * Calls `add_path_recursively` only for the `Recursive` paths.
  * Calls `add_path` for both types.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -37,8 +37,18 @@ pub struct RootedPath {
     pub path: StorePath,
 }
 
+/// A path that was accessed during instantiation and should be watched
+/// for changes.
+#[derive(Debug)]
+pub enum ReferencedPath {
+    /// Should watch the whole directory (if the path is a directory).
+    Recursive(PathBuf),
+    /// Just watch path that was accessed.
+    NotRecursive(PathBuf),
+}
+
 struct InstantiateOutput {
-    referenced_paths: Vec<PathBuf>,
+    referenced_paths: Vec<ReferencedPath>,
     output: RootedDrv,
 }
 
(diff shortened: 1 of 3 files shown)


hamishmack:
@curiousleo, does this look ok?

Profpatsch:
It looks like the patch does not type-check currently (see travis CI log)

src/builder.rs
```diff
@@ -37,8 +37,48 @@ pub struct RootedPath {
...
 
+/// A path that was accessed during instantiation and should be watched
+/// for changes.
+#[derive(Debug)]
+pub enum ReferencedPath {
+    /// Should watch the whole directory (if the path is a directory).
+    Recursive(PathBuf),
+    /// Just watch path that was accessed.
+    NotRecursive(PathBuf),
+}
```
curiousleo, line 48: Would it not be simpler to track recursively and non-recursively watched paths in separate `Vec<PathBuf>`s?

codygman:
The problem this fixes (creating tons of watchers) just crashed my coworkers computer and he had to hard reboot. Just going to `use_nix` for now instead of lorri to work around it.

I don't know much Rust, only Haskell, but I guess I'll start learning to try and make the suggested changes in this PR.

Any recommended resources?

curiousleo:
@codygman that's great, thanks for the offer!

The Rust compiler usually produces helpful error messages. In my workflow, I always have `cargo watch` (the Nixpkgs name is `cargo-watch`) running in a terminal next to my editor, which is basically `stack build --file-watch` for Rust.

https://doc.rust-lang.org/stable/rust-by-example/ has lots of examples for Rust syntax and idioms.
https://doc.rust-lang.org/book/ has more details.

symphorien:
If there is still interest in this PR, could you reopen it on https://github.com/nix-community/lorri ?

Profpatsch:
@symphorien heroically re-opened it in https://github.com/nix-community/lorri/pull/34, PTAL

Issue #351, opened 2020-03-16
lorri crashes on `lorri watch` from emacs shell.


mwillsey:
**Describe the bug**

lorri crashes on `lorri watch` from emacs shell. Crash log seems to indicate it's due to incorrect terminal capability handing.

**To Reproduce**

1. Open a (bash) shell in emacs
2. Navigate to a directory with a shell.nix and a .envrc
3. Run `lorri watch`

**Metadata**

```console
$ lorri info --shell-file shell.nix 
lorri version: 1.0
gc roots exist, shell_gc_root: "/home/mwillsey/.cache/lorri/gc_roots/b248d494210f65e15ff407b42e045d41/gc_root/shell_gc_root"
```

```
$ uname -a
Linux nixos 5.4.24 #1-NixOS SMP Thu Mar 5 15:43:52 UTC 2020 x86_64 GNU/Linux
```

Crash report

```
name = 'lorri'
operating_system = 'unix:Unknown'
crate_version = '1.0.0'
explanation = '''
Panic occurred in file '/build/lorri-1.0-vendor.tar.gz/slog/src/lib.rs' at line 1914
'''
cause = 'slog::Fuse Drain: Custom { kind: Other, error: "term error: operation not supported by the terminal" }'
method = 'Panic'
backtrace = '''

   0: 0x5573292e4aa0 - std::panicking::try::do_call::h16e5a52efd511179
   1: 0x55732951b3fa - __rust_maybe_catch_panic
   2: 0x55732934e697 - core::ops::function::FnOnce::call_once{{vtable.shim}}::h53598dca4ddea8c8
   3: 0x55732950f1ef - <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once::he9dcb7073be13c65
   4: 0x5573294fce30 - std::sys_common::thread::start_thread::h9ce8f903e84fe8e9
   5: 0x55732950e866 - std::sys::unix::thread::Thread::new::thread_start::h5e894f0575174d9e
   6: 0x7f8abbcd5edd - start_thread
   7: 0x7f8abbbeca4f - __clone
   8:        0x0 - <unknown>'''
```

Profpatsch:
cc @curiousleo

curiousleo:
[...]

PR #359, opened 2020-03-23
Update slog_term to 2.5.0 and human-panic to 1.0.3

curiousleo:
Relates to #351.

## Overview

Updates `slog_term` to 2.5.0. This ought to fix https://github.com/target/lorri/issues/351 but doesn't. More context: https://github.com/slog-rs/slog/issues/160 and https://github.com/slog-rs/term/commit/93587a3533783bdf269b237ca4a441ee7b24f0f7.

## Checklist


Cargo.lock
```diff
@@ -1311,7 +1345,9 @@ dependencies = [
 "checksum cloudabi 0.0.3 (registry+https://github.com/rust-lang/crates.io-index)" = "ddfc5b9aa5d4507acaf872de71051dfd0e309860e88966e1051e462a077aac4f"
 "checksum constant_time_eq 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)" = "995a44c877f9212528ccc74b21a232f66ad69001e40ede5bcee2ac9ef2657120"
 "checksum crossbeam-channel 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)" = "c8ec7fcd21571dc78f96cc96243cab8d8f035247c3efd16c687be154c3fa9efa"
+"checksum crossbeam-channel 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)" = "cced8691919c02aac3cb0a1bc2e9b73d89e832bf9a06fc579d4e71b68a2da061"
 "checksum crossbeam-utils 0.6.6 (registry+https://github.com/rust-lang/crates.io-index)" = "04973fa96e96579258a5091af6003abde64af786b860f18622b82e026cca60e6"
+"checksum crossbeam-utils 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)" = "c3c7c73a2d1e9fc0886a08b93e98eb643461230d5f1925e4036204d5f2e261a8"
```
Profpatsch: Hm, now we have lots of duplicates here. Can we also update the other dependency to keep the transitive graph small?

curiousleo: Unfortunately, `notify v5.0.0-pre.1` depends on `crossbeam-channel v0.3` so we can't use `crossbeam-channel v0.4.2` throughout.

We can't update `notify` easily because there are breaking changes between the version we're currently using and the next.

```
[nix-shell:~/Code/lorri]$ cargo tree -p crossbeam-utils:0.7.2 -i
crossbeam-utils v0.7.2
└── crossbeam-channel v0.4.2
    └── slog-async v2.5.0
        └── lorri v1.0.0 (/home/leo/Code/lorri)

[nix-shell:~/Code/lorri]$ cargo tree -p crossbeam-utils:0.6.6 -i
crossbeam-utils v0.6.6
└── crossbeam-channel v0.3.9
    ├── lorri v1.0.0 (/home/leo/Code/lorri)
    └── notify v5.0.0-pre.1
        └── lorri v1.0.0 (/home/leo/Code/lorri) (*)
```

However, overall this PR reduces the size of the dependencies specification by about 100 lines and removes a few transitive dependencies that are no longer required by newer versions of our dependencies (pretty amazing actually: newer versions reduce dependencies instead of adding new ones!)

Profpatsch: Thanks for the `cargo tree` example, I will have to remember that feature.

> pretty amazing actually: newer versions reduce dependencies instead of adding new ones!

I would sure hope the rust ecosystem settles down to a tight dependency tree with less breaking changes as time goes on, as opposed to the hot mess (high entropy!) it is right now :P

curiousleo: Responding to this thread and partially to https://github.com/target/lorri/issues/351#issuecomment-603467981: when we adopted `slog`, we sort of went all-in. In this PR, I just removed `slog_async` and replaced it by a global mutex. I think that's fine because logging is unlikely to be a bottleneck. If it ever becomes one, we'll just add `slog_async` back in.

This also removes the need for a second `crossbeam-channel`:

```
[nix-shell:~/Code/lorri]$ cargo tree -p crossbeam-channel -i
crossbeam-channel v0.3.9
├── lorri v1.0.0 (/home/leo/Code/lorri)
└── notify v5.0.0-pre.1
    └── lorri v1.0.0 (/home/leo/Code/lorri) (*)
```

Profpatsch:
The MacOS build timed out while it was uploading the final cache …

Profpatsch:
While we are at it I bumped the human-panc version to the official release.

Profpatsch:
Okay, this is not going to fly. :( 
We need to switch to run `cachix push` in nix post-update-hooks, so that new derivations are uploaded to the cache as soon as they are built.

I have the code for this set up in another project, I’ll try to add it to this PR soon(ish).

Profpatsch:
I’ll go ahead and merge this, since 3/4 builds are green and the macos one just timed out very shortly before sucessful completion (when uploading the new cache).

PR #363, opened 2020-03-26
safely handle spaces in PATH

### LINKED ISSUE #18: lorri (or direnv?) isn't clean about spaces in PATH
```
~> cd opt/lorri/example/
direnv: loading .envrc
/bin/bash: line 66: /Users/xxx/Library/Caches/com.github.target.lorri.lorri.lorri/example/gc_root/build-0: No such file or directory
direnv: error exit status 1
~/o/l/example> cd..
fish: Unknown command cd..
direnv: loading .envrc
/bin/bash: line 29: Fusion.app/Contents/Public:/Users/xx/.nix-profile/bin:/Users/xx/.nix-profile/lib/kde4/libexec:/nix/var/nix/profiles/default/bin:/nix/var/nix/profiles/default:/nix/var/nix/profiles/default/lib/kde4/libexec:/Applications/kitty.app/Contents/MacOS:/Users/xx/bin: No such file or directory
direnv: error exit status 1
```

christian-blades-cb:
Closes #18.

## Overview

Quoting the value of variable in the prepend function prevents `eval` from treating everything after a space as another command to evaluate.

```
❯ echo $PATH
/Users/blades/.nix-profile/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/Library/TeX/texbin

❯ lorri shell
lorri: building environment......... done
/Users/blades/Library/Caches/com.github.target.lorri.lorri.lorri/cas/fe2381e5f5b538e916a10973b0a6dea1: line 61: Fusion.app/Contents/Public:/Library/TeX/texbin: No such file or directory

❯ ./target/debug/lorri shell
lorri: building environment......... done
```

## Checklist

~- [ ] Updated the documentation (code documentation, command help, ...)~

### DIFF
--- a/tests/integration/envrc.rs
+++ b/tests/integration/envrc.rs
@@ -76,3 +76,21 @@ fn v2_gopath_previously_unset() {
 
     assert_eq!(env.get_env("GOPATH"), DirenvValue::Value("BAR"));
 }
+
+#[test]
+fn bug18_path_with_spaces() {
+    let exemplar = "/mnt/c/Program Files (x86)/QuickTime/QTSystem";
+    let path = env::var("PATH")
+        .map(|x| format!("{}:{}", x, exemplar))
+        .unwrap();
+    let env = EnvrcTestCase::new()
+        .ambient_env("PATH", &path)
+        .project_env(ProjectEnvBuilderV1::new().set("PATH", "/foo/bar"))
+        .unwrap()
+        .get_direnv_variables();
+
+    assert_eq!(
+        env.get_env("PATH"),
+        DirenvValue::Value(&format!("/foo/bar:{}", &path))
+    );
+}
--- a/release.nix
+++ b/release.nix
@@ -6,6 +6,12 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 517;
+        changes = ''
+          - Fix issue with spaces in PATH entries
+        '';
+      }
+      {
         version = 510;
         changes = ''
           - The shell.nix template used by `lorri init` was changed to take
--- a/src/ops/direnv/envrc.bash
+++ b/src/ops/direnv/envrc.bash
@@ -55,7 +55,7 @@ function prepend() {
     # reference, plus the current (updated on the export) contents.
     # however, exclude the ${separator} unless ${original} starts
     # with a value
-    eval "$varname=${!varname}${original:+${separator}${original}}"
+    eval "$varname=\"${!varname}${original:+${separator}${original}}\""
 }
 
 function append() {


tests/shell/main.rs
```diff
@@ -27,6 +27,26 @@ fn loads_env() {
     );
 }
 
+#[test]
+fn path_with_spaces() {
+    let exemplar = "/opt/hello world/bin/:/mnt/c/Program Files (x86)/QuickTime/QTSystem:";
+    let new_path = std::env::var("PATH")
+        .map(|p| format!("{}:{}", p, exemplar))
+        .unwrap_or_else(|_| String::from(exemplar));
+    std::env::set_var("PATH", new_path);
```
curiousleo: If I'm understanding it correctly, then what's important in this test is that the environment is set for the command (on line 39), right?

If so, my suggestion would be to remove the side-effect of changing this processes' environment and set the environment of the sub-process launched on line 39 instead, see https://doc.rust-lang.org/std/process/struct.Command.html#method.env.

I think it is then not necessary to manually query the current `"PATH"` and prepend it to the example. But I'm not 100% sure this works.

Just to be sure: can you confirm that this test fails without the changes to `envrc.bash`?

christian-blades-cb: You're correct, the PATH var is the important bit of the test.

I'll take a stab at modifying the environment of just the process here. I think I'll still need to append my exemplars to the current path so that `echo` works.

christian-blades-cb: Works without appending to the existing PATH 🙏 

Undo-ing my changes to `envrc.bash` breaks the test (and also `trivial_v1` and `trivial_v2` since they reference my current `PATH` which includes spaces)

christian-blades-cb: Which, now that I'm looking at it again, makes me realize that the test really belongs in `envrc.rs`, I'll rewrite it there.

tests/integration/envrc.rs
```diff
@@ -76,3 +76,21 @@ fn v2_gopath_previously_unset() {
 
     assert_eq!(env.get_env("GOPATH"), DirenvValue::Value("BAR"));
 }
+
+#[test]
+fn bug18_path_with_spaces() {
```
curiousleo: What does the "18" refer to here?

tests/integration/envrc.rs
```diff
@@ -76,3 +76,21 @@ fn v2_gopath_previously_unset() {
...
+    let exemplar = "/mnt/c/Program Files (x86)/QuickTime/QTSystem";
+    let path = env::var("PATH")
+        .map(|x| format!("{}:{}", x, exemplar))
+        .unwrap();
+    let env = EnvrcTestCase::new()
+        .ambient_env("PATH", &path)
```
curiousleo:
```suggestion
        .ambient_env("PATH", &[exemplar])
```

Would this also work? That way, there is no need to query the actual environment variable PATH with `env::var("PATH")`. If we can avoid pulling parts of the actual environment into our tests, it makes the tests more resilient against changes in the environment (personal machines vs. CI, for example).

curiousleo: That's what I meant to suggest in https://github.com/target/lorri/pull/363#discussion_r399084948, but perhaps I didn't make it very clear then.

tests/integration/envrc.rs
christian-blades-cb: There seemed to be a naming convention for regression tests, so this refers to issue #18

tests/integration/envrc.rs
christian-blades-cb: It won't accept a slice, and without appending to the existing PATH I get
```

Profpatsch:
Many thanks for fixing this bug!

PR #368, opened 2020-04-04
Watch refactor path adding logic

Profpatsch:
The refactor I mentioned in https://github.com/target/lorri/pull/355#issuecomment-602318309

Splits the recursive enumeration of files from the actual adding to the watcher.

cc @hamishmack @curiousleo

### DIFF
--- a/src/watch.rs
+++ b/src/watch.rs
@@ -5,6 +5,7 @@ use crate::NixFile;
 use crossbeam_channel as chan;
 use notify::event::ModifyKind;
 use notify::{EventKind, RecommendedWatcher, RecursiveMode, Watcher};
+use slog_scope::trace;
 use slog_scope::{debug, info};
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
@@ -109,14 +110,17 @@ impl Watch {
     /// Extend the watch list with an additional list of paths.
     /// Note: Watch maintains a list of already watched paths, and
     /// will not add duplicates.
-    pub fn extend(&mut self, paths: &[PathBuf]) -> Result<(), notify::Error> {
+    pub fn extend(&mut self, paths: Vec<PathBuf>) -> Result<(), notify::Error> {
         for path in paths {
-            self.add_path(&path)?;
-            if path.is_dir() {
-                self.add_path_recursively(&path)?;
+            let recursive_paths = walk_path_topo(path)?;
+            for p in recursive_paths {
+                if p.canonicalize()?.starts_with(Path::new("/nix/store")) {
+                    trace!("Skipping watching nix store path {}", p.display());
+                    continue;
+                }
+                self.add_path(p)?;
             }
         }
-
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


src/watch.rs
```diff
@@ -195,6 +201,60 @@ impl Watch {
...
+    let mut files = vec![];
+    for entry in std::fs::read_dir(dir)? {
+        let entry = entry?;
+        if entry.file_type()?.is_dir() {
+            dirs.push(entry.path())
+        } else {
+            files.push(entry.path())
+        }
+    }
+    Ok((dirs, files))
```
curiousleo, line 217: Shouldn't something like this do the trick?
```
    Ok(std::fs::read_dir(dir)?.partition(|e| e.file_type()?.is_dir()))
```

Ah no, can't have that `file_type()?` in there ... that sucks.

src/watch.rs
```diff
@@ -195,6 +201,60 @@ impl Watch {
...
+}
+
+/// List all children of the given path.
+/// Recurses into directories.
```
curiousleo, line 221: ... but avoids being recursive :) Is there a reason (bechmarked performance or something like that) that you went for full-blown imperative style here? I suspect that `walk_path_topo` could call itself, simplifying the code a bit.

src/watch.rs
Profpatsch: yeah, in Rust a general rule of thumb I’ve found is that side-effecting stuff is better in imperative loops.

src/watch.rs
Profpatsch: It felt a lot easier to do than going for a more functional approach (especially for topological sort).
Also it doesn’t depend on the call stack, which is usually the better approach for imperative languages (though I don’t think we would have directories that deep, but you never know. There’s always that *one* user).

curiousleo: Here's an alternative, functional way to write the same function.

```rust
fn walk_path_topo(path: PathBuf) -> Result<Vec<PathBuf>, std::io::Error> { 
    let p = vec![path.clone()];
    if !path.is_dir() {    
        return Ok(p);
    }                                                                  
    let (dirs, files) = list_dir(&path)?;    
    Ok(p 
        .into_iter()
        .chain(files.into_iter())
        .chain(
            dirs.into_iter()
                .map(walk_path_topo)
                .collect::<Result<Vec<Vec<PathBuf>>, _>>()?
                .into_iter()
                .flatten(),
        )
        .collect())
}
```

src/watch.rs
```diff
@@ -109,17 +115,35 @@ impl Watch {
     /// Extend the watch list with an additional list of paths.
     /// Note: Watch maintains a list of already watched paths, and
     /// will not add duplicates.
-    pub fn extend(&mut self, paths: &[PathBuf]) -> Result<(), notify::Error> {
+    pub fn extend(&mut self, paths: Vec<PathBuf>) -> Result<(), notify::Error> {
```
nyarly, line 118:
```suggestion
    pub fn extend(&mut self, paths: IntoIterator<Item=PathBuf>) -> Result<(), notify::Error> {
```
I tend to abstract to a trait when I change types like this

src/watch.rs
Profpatsch: It’s self recursive though, so the stack might blow for edge cases (very deep paths). Amazingly the order is the exact same, but I have no idea about the performance of the iterator chain, whereas with the internally mutable code I can see what is going on (mutable appending to a deque and popping from the beginning). Not that matters mutch, since we are doing I/O at any step.

I think I’ll keep the internally mutating code, since it’s fairly straightforward.

src/watch.rs
Profpatsch: We’re not exposing it as library, so abstractions like that are not really needed (they only complicated the code somewhat).

src/watch.rs
Profpatsch: But it’s great to see a use of `chain`, I will have to look at how that is implemented internally.

PR #370, opened 2020-04-07
Start an Associated Projects README section

### LINKED ISSUE #40: `lorri pin` command
@aspiwack writes:
> Pinning nixpkgs is very underused. Mostly because it's a rather painful process (one has to find the commit reference for the current nixpkgs then remember the code snippet to copy/paste in).
> 
> So I imagine that lorri could have a command to do all these things for me.
>
>----
>
> Here is a straw-man proposal:
>
> Pin the project to the current nixpkgs version:
>
>```
> $ lorri pin
>```
>
> Pin the project to the latest nixpkgs-stable:
>
>```
> $ lorri pin nixpkgs-stable
>```
>
> Update the pinned version to the current nixpkgs:
>
>```
> $ lorri pin --update
>```
>
> I'm not sure what it should look like nix-expression-wise. Certainly the pinning action should be in file checked into the version control system. Possibly by modifying the shell.nix file. Or possibly by making a lock.nix file under lorri's control, which you would require the user to call from shell.nix. Something to that effect.

@zimbatm writes:
> for some prior art see:
>
> * https://github.com/nmattia/niv/
> * https://github.com/zimbatm/nix-path

@shlevy writes:
> If we do this it would be good to coordinate with flakes if relevant.

nyarly:
Adds a note and link to direnv, Nix and niv.

Closes #40

Also provides a home in the docs for further links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,22 @@ help you get started.
 This section needs to be fleshed out more
 ([#244](https://github.com/target/lorri/issues/244)).
 
+## Associated Projects
+
+lorri embodies a Unix philosophy of doing one thing well.
+As a result, it works very well with other tools to provide a powerful
+and streamlined developement experience.
+
+[direnv](https://github.com/) is lorri's chief collaborator. From its website:
+"It augments existing shells with a new feature that can load and unload
+environment variables depending on the current directory."
+
+As a foundation, lorri relies on [Nix and Nixpkgs](https://nixos.org/nix/)
+to install and manage software packages.
+
+For pinning versions of software during development,
+[niv](https://github.com/nmattia/niv) is very helpful.
+
 ---
 
 ## Support & Questions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,6 +142,22 @@ help you get started.
 This section needs to be fleshed out more
 ([#244](https://github.com/target/lorri/issues/244)).
 
+## Associated Projects
+
+lorri embodies a Unix philosophy of doing one thing well.
+As a result, it works very well with other tools to provide a powerful
+and streamlined developement experience.
```
curiousleo, line 149:
```suggestion
and streamlined development experience.
```

README.md
```diff
@@ -142,6 +142,22 @@ help you get started.
...
+
+[direnv](https://github.com/) is lorri's chief collaborator. From its website:
+"It augments existing shells with a new feature that can load and unload
+environment variables depending on the current directory."
```
curiousleo, line 153:
```suggestion
> [direnv] augments existing shells with a new feature that can load and unload
> environment variables depending on the current directory.
```

Just looks prettier when rendered IMO :)

PR #373, opened 2020-04-11
daemon: Make it possible to pass `substituters` and `builders`

Profpatsch:
Adds the ability to input additional nix arguments through to the nix
commands. For our sanity and so that we don’t have to replicate the
whole nix API, they are passed as json string on the command line.

However, I don’t know how an exhaustive API would look like (nix has a
lot of configuration options and their semantics is pretty weird in
places), so I opted to only add `--substituters` and `--builders`. We
even gain a nicer interface by using an optional list.

I’m not yet positive on how to document the JSON interface more
nicely, maybe we do want to expose all options as command line flags,
but I really don’t know how that would look without leading to a mess.

The changes are rather simple and mostly mechanical.

As a followup, we should add the same flag to `watch` and `shell`.

cc @curiousleo @nyarly 

## Checklist

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -47,7 +47,7 @@ pub enum Command {
 
     /// Start the multi-project daemon. Replaces `lorri watch`
     #[structopt(name = "daemon")]
-    Daemon,
+    Daemon(DaemonOptions),
 
     /// Upgrade Lorri
     #[structopt(name = "self-upgrade", alias = "self-update")]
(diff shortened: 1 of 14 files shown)


nomeata:
There are more options one might want to pass. In particular `--option tarball-ttl 1000000` is super important when working on a train or plane.

Profpatsch:
Whoo, time to merge this then.

src/cli.rs
```diff
@@ -118,6 +118,33 @@ pub struct WatchOptions {
...
+    /// {
+    ///   "builders": <optional list of string>,
+    ///   "substituters": <optional list of string>
+    /// }
+    pub extra_nix_options: Option<NixOptions>,
+}
+
+/// The nix options we can parse as json string
+#[derive(Deserialize, Debug)]
+// ATTN: If you modify this,
```
curiousleo, line 139: "ATTN"?

Profpatsch: “attention”

Similar to TODO and XXX, but if you want to get a developer to hopefully read the comment before modifying the code :)

PR #376, opened 2020-04-15
ops: banish underscores in cli commands

Profpatsch:
Either we do `-` or `_`, and we already use `-` for all flags and
`self-upgrade`. Consistency is king for good UX.

This is obviously a breaking change, but these are `internal` and haven’t seen an official release so far (plus we added the `internal`-as-a-subcommand thing recently anyway).

cc @curiousleo @nyarly

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,6 +6,14 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 534;
+        changes = ''
+          Rename `lorri internal` commands for consistency:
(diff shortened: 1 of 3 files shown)


Profpatsch:
Look at them github actions actioning. cc @lovesegfault

Profpatsch:
Meh, the Github action setup is not extremely good yet. But it’s a start.

Profpatsch:
Merging, since the travis build was green before I pushed a trivial change to see the Github Action output.

Issue #383, opened 2020-04-21
Lorri broken prompt on nested shells, bash vs bashInteractive and its cohort of issues


deliciouslytyped:
See also https://github.com/NixOS/nixpkgs/issues/59209#issuecomment-617103619 , running shells in direnv+Lorri on NixOS runs into the usual bash vs bashInteractive issues with broken prompts.

It can be worked around by adding bashInteractive to the mkShell.

kaleocheng:
I also got this issue on NixOS

.envrc:
```
eval "$(lorri direnv)"
layout python2
```

shell.nix
```
let
  pkgs = import <nixpkgs> {};
in
pkgs.mkShell {
  shellHook = ''
    unset SOURCE_DATE_EPOCH
  '';
}
```


when I run a nested bash inside the direnv+Lorri bash, I got a broken prompt:

```
$ bash 
complete: command not found
bind: command not found
bind: command not found
bash: set: vi: invalid option name
\[\]\[\] impure \[\]\[\]\[\]\[\]\[\] ~ \[\]\[\]\[\]\[\]\[\] \[\]\[\]\[\]\[\]\[\] … \[\]\[\]\[\]\[\]\[\] $
```

@deliciouslytyped  really appreciate if you could show more details about the workaround

Profpatsch:
```
let
  pkgs = import <nixpkgs> {};
in
pkgs.mkShell {
  shellHook = ''
    unset SOURCE_DATE_EPOCH
  '';
  buildInputs = [ pkgs.bashInteractive ];
}
```

should work around this.

Actually I think we should add that to the default `shell.nix` we generate, since it’s a known problem.

PR #392, opened 2020-05-06
src/trivial-shell.nix: add bashInteractive

Profpatsch:
For bash users, `mkShell` overrides their path with a non-interactive
shell, which breaks their shells.

Adding `bashInteractive` is a workaround, so we should recommend that
and add it via `lorri init`.

Closes https://github.com/target/lorri/issues/383

cc @deliciouslytyped


deliciouslytyped:
Probably fine, thanks.

PR #393, opened 2020-05-08
(trivial) src/error.rs: use Debug for std::io::Error

Profpatsch:


Profpatsch:
> I assume you came across an example where `Display` didn't show enough information and `Debug` does? I think that if you have an example like that, the PR description would be a good place to put it to motivate the PR.

yeah, a user was reporting an issue in the IRC channel, and the error message was missing any context.

Since the `UnrecoverableError`s are our “lorri is going to crash now” path, we should be as verbose as possible.

PR #395, opened 2020-05-10
logged-evaluation: add IN_LORRI_SHELL

### LINKED ISSUE #372: Add `IN_LORRI_SHELL` environment variable
**Feature description**

It would be nice if `lorri` set some kind of environment variable to indicate that you aren't just in a Nix shell, but specifically one provided by `lorri`. Something like `IN_LORRI_SHELL`.

**Target users**

- End users who want to customize their shell prompt to indicate that they're in a `lorri` shell
- The `lorri` developers, so they can prevent me from invoking `lorri shell` over and over in a nested fashion 🙂 I'm thinking it should warn like `tmux` does:

    ```
    sessions should be nested with care, unset $TMUX to force
    ```

cole-h:
This will allow users to differentiate between a `nix-shell` invocation
and `lorri shell` invocation. One potential usecase is customizing the
prompt based on whether or not one is in a lorri shell.

Closes #372.

## Overview

* I added `IN_LORRI_SHELL` right below `IN_NIX_SHELL` in `logged-evaluation.nix`.
  * I opted not to implement the suggested nested check because, when combined with an `.envrc`, `lorri shell` would always be disallowed unless `IN_LORRI_SHELL` was unset. This can be easily added back in, however; let me know.
* ~~I fixed the warnings-turned-errors that cropped up when running the `ci` script in a `nix-shell`. I can split this out to another PR, if you'd prefer.~~ I've moved this commit to #396, which also addresses the GitHub Actions CI failures.

## Checklist


tests/integration/bug97_varmap_leak.rs
```diff
@@ -32,6 +32,7 @@ fn bug97_varmap_leak() {
         "origExtraClosure",
         // Nix-set variables
         "IN_NIX_SHELL",
+        "IN_LORRI_SHELL",
```
curiousleo, line 35: Nit: this is listed under "Nix-set variables" although it is really set by lorri. I suggest creating a new "section" here, perhaps "lorri-set variables".

cole-h: I noticed this too haha. I'll add another section along with the integration test.

cole-h:
OK, I added an integration test and a new section for `bug97`. I couldn't find a good place to put it (since it doesn't really involve the `.envrc`, I was hesitant to put it in `envrc.rs`), so I moved all the stuff present in `direnv.rs` into `direnvtestcase.rs` and repurposed `direnv.rs` into a file for tests like mine (I really didn't want to add another file for something so trivial, so I just took over an existing one).

cole-h:
Great idea! I've done just that.

release.nix
```diff
@@ -5,6 +5,13 @@
   changelog = {
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
+      {
+        version = 548;
+        changes = ''
+          Add $IN_LORRI_SHELL environment variable to allow differentiation
+          between `nix-shell` and `lorri shell`.
```
Profpatsch, line 12: We should mention that it carries the path to the evaluated shell file as value.

src/logged-evaluation.nix
```diff
@@ -129,9 +129,14 @@ let
...
             # Export IN_NIX_SHELL to trick various Nix tooling to export
             # shell-friendly variables
-
             export IN_NIX_SHELL=impure
 
+            ${if shellSrc == null then "" else ''
+              # Export IN_LORRI_SHELL to differentiate between `lorri shell` and
+              # `nix-shell`
+              export IN_LORRI_SHELL=${shellSrc}
+            ''}
```
Profpatsch, line 138: The `null` conditional looks strange to me; @curiousleo do we still require this switch for the services? I think we only ever had a WIP for services, do we want to keep them around in the codebase?

cole-h: If `shellSrc ? null` turns out to be unnecessary (pending @curiousleo's response), should I just remove `? null` from the argset?

curiousleo: I suggest we merge #405 first and rebase this PR on top of it. I've filed #404 for follow-up work to remove complexity due to the incomplete `lorri services` implementation.

Profpatsch:
I merged #405, so if you rebase we can just include it unconditionally.

src/logged-evaluation.nix
```diff
@@ -122,9 +122,12 @@ let
 
             # Export IN_NIX_SHELL to trick various Nix tooling to export
             # shell-friendly variables
-
             export IN_NIX_SHELL=impure
 
+            # Export IN_LORRI_SHELL to differentiate between `lorri shell` and
+            # `nix-shell`
+            export IN_LORRI_SHELL=${src}
```
Profpatsch, line 129: This needs to be escaped via `lib.escapeShellArg`. I think it’s gonna be a lot of work importing `lib` into this file, so let’s just inline the definition from `nixpkgs/lib/strings.nix` into this file:

```
  escapeShellArg = arg: "'${builtins.replaceStrings ["'"] ["'\\''"] (toString arg)}'";
```

cole-h: What about spaces? Should we worry about those? As it stands, having a folder `asd fg'h` will break e.g. `stat $IN_LORRI_SHELL`.

Profpatsch: `escapeShellArg` will do the right thing for both spaces and single quotes. It wraps everything in single quotes, taking care of spaces, and then it leaves the single quote every time it encounters one, to prevent escaping issues.

`ab c'de` -> `'ab c'\''de'`

Bash implicitly concatenates two strings if they have no spaces between each other, so both single-quoted strings are merged into one string.

Profpatsch:
Thanks!

PR #396, opened 2020-05-10
Fix GH Actions CI

cole-h:
## Overview

After submitting my PR implementing `$IN_LORRI_SHELL`, I noticed a lot of CI was failing. Thus, I decided to spend a little bit to fix up what I could. CI should pass with flying colors, now.

## Checklist

---

~~EDIT: About to sleep, but it looks like that Travis failure might be fixable after all. I'll check in the morning.~~ ~~I can't reproduce the warning/error locally... I guess I'll take this chance to see if I can't move us entirely to GH Actions.~~ nvm, I needed to use Rust 1.41.0 to see the issue. Latest stable apparently doesn't have this problem.


Profpatsch:
Super!

Now all we need for feature parity with travis is a macOS runner as far as I can see. Unfortunately, Github Actions does not provide macOS builders on its free tier (reasonable), and the paid version is super expensive (8¢/*minute*). I wonder how hard it is to install the runner on an old macbook we have flying around.

nix/update-dependencies.sh
```diff
@@ -4,7 +4,7 @@
 set -euo pipefail
 
 # lorri should always build with the current NixOS stable branch.
-channel='nixos-19.09'
```
curiousleo, line 7: Currently, lorri "officially" supports 19.09 - see https://github.com/target/lorri/blob/master/MAINTAINERS.md#publishing-a-release-on-nixpkgs. Switching CI to 20.03 means that we pull in a newer version of Rust. This also means that we may start using Rust features that are not supported by the Rust version shipped with 19.09.

As such, this change is more than just a fix to CI; it implies soft-deprecating 19.09. If there is a way to separate out fixing CI from deprecating 19.09, then I would be in favour of splitting the PR and then having a separate discussion about which NixOS versions we want to support.

cole-h: The main reason I soft-deprecated `nixos-19.09` is because it would consume a huge chunk of time just building the overlayed `nixpkgs-fmt` (easily 10 minutes just to enter the shell for the first time). One way this could reasonably be solved is by maybe utilizing the [`nixpkgs-mozilla` overlay](https://github.com/mozilla/nixpkgs-mozilla/) to pin the Rust packages to 1.37.0 (the version in `nixos-19.09`) and having jobs for both 1.41.0 and 1.37.0 utilizing that overlay. What do you think about that?

cole-h:
@Profpatsch I've actually been working on bringing GH Actions on-par with Travis on my branch here: https://github.com/cole-h/lorri/pull/3 (ignore the super-descriptive name).

I don't think macOS builders actually cost money for public Actions. As you'll see in my above link above, I've added macOS to the matrix (and tested the macOS stuff many, many times) and still have a grand total of 0 minutes used on my account and not a dime owed on the Billing page.

I think the confusion comes from the fact that the following sentence is not very well-advertised: "GitHub Actions usage is free for public repositories." (from [here](https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions)) This might come from the fact that I am considered "Pro" (yay, being a student!) and a user account instead of an organization like Target, however.

nix/update-dependencies.sh
curiousleo: @cole-h - thanks for suggesting an alternative with the Rust overlay! I believe that what you're saying would work. It would also somewhat complicate the setup.

My main point was mostly that the decision to "effectively" deprecate 19.09 should be taken consciously. I am personally quite happy to drop support for 19.09 now that 20.03 is out, which would allow us to go ahead without the Rust overlay.

@grahamc @Profpatsch @nyarly, are you okay with us dropping support for 19.09 and move to 20.03 as the officially supported version (by way of running CI using 20.03)?

If yes, let's change all occurrences of "19.09" to "20.03" in https://github.com/target/lorri/blob/master/MAINTAINERS.md too (and also in the README in case we mention 19.09 somewhere there).

curiousleo: Proposed docs change: https://github.com/target/lorri/pull/399

Profpatsch: > @grahamc @Profpatsch @nyarly, are you okay with us dropping support for 19.09 and move to 20.03 as the officially supported version (by way of running CI using 20.03)?

Let’s keep support for 19.09 around until 20.09 is released. We can use the pinned mozilla rustpkgs overlay.

Profpatsch:
> EDIT2: Update: still no minutes used and no money due in the Billing page. I think it should be safe?

Gonna push your branch and see what happens.

nix/update-dependencies.sh
curiousleo: I've filed https://github.com/target/lorri/issues/406.

PR #399, opened 2020-05-14
CI runs against nixos-20.03 now

curiousleo:
Follow-up to https://github.com/target/lorri/pull/396#discussion_r422865040.

## Overview

Document the fact that CI runs against nixos-20.03 now.

## Checklist

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -66,9 +66,11 @@ To cut a new release:
 
 ## Publishing a release on [nixpkgs][]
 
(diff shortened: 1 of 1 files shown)


Profpatsch:
Let’s keep 19.09 supported at least for half a year more. We don’t have much reason to drop support, we can just pin the build to 19.09 rust.

curiousleo:
@Profpatsch `MAINTAINERS.md` still needs to be updated I think, since it currently talks _only_ about 19.09, so it is out of date.

Profpatsch:
Done in https://github.com/target/lorri/pull/438

PR #400, opened 2020-05-14
Account for .zshenv or /etc/zshrc setting ZDOTDIR

### LINKED ISSUE #398: `lorri shell` doesn't support changing the zsh configuration directory at startup
**Describe the bug**

lorri [looks up the `ZDOTDIR` environment variable](https://github.com/target/lorri/blob/make-commands-use-hyphens/src/ops/start_user_shell.rs#L72) and stores its value as [`ZDOTDIR_OR_HOME`](https://github.com/target/lorri/blob/make-commands-use-hyphens/src/ops/start_user_shell.rs#L71). In typical cases `ZDOTDIR` is a *shell variable*, not an environment variable, so `ZDOTDIR_OR_HOME` will be set to the value of `$HOME`. 

Then lorri [sources the `.zshenv` file](https://github.com/target/lorri/blob/make-commands-use-hyphens/src/ops/start_user_shell.rs#L63). The code in `.zshenv` might change the value of `ZDOTDIR` shell variable. But then lorri [uses the old `ZDOTDIR_OR_HOME` location](https://github.com/target/lorri/blob/make-commands-use-hyphens/src/ops/start_user_shell.rs#L65), not the new `ZDOTDIR` location. So if `ZDOTDIR` is changed in `.zshenv`, that change will not be observed.

**To Reproduce**

1. `echo ZDOTDIR="$HOME"/.config/zsh > "$HOME"/.zshenv"`
2. `echo HISTFILE="$HOME/.config/zsh/zsh_history >> "$HOME"/.config/zsh/.zshrc`
3. Open a new shell
5. `echo $HISTFILE` 
See that it shows the adjusted `zsh_history` path.
3. `lorri shell` 
4. `echo $HISTFILE`
See that it's empty.

**Ideas**

It might be enough to simply load the new `ZDOTDIR` after sourcing each file

```zsh
test -f "$ZDOTDIR_OR_HOME/.zshenv" && . "$ZDOTDIR_OR_HOME/.zshenv"
ZDOTDIR_OR_HOME="${ZDOTDIR:-${HOME}}"
test -f "/etc/zshrc"               && . "/etc/zshrc"
ZDOTDIR_OR_HOME="${ZDOTDIR:-${HOME}}"
test -f "$ZDOTDIR_OR_HOME/.zshrc"  && . "$ZDOTDIR_OR_HOME/.zshrc"
```

but I haven't tested it.

**Metadata**

```console
$  lorri --version
lorri 1.0.0
```

```console
$ uname -a
Linux beta 5.4.35 #1-NixOS SMP Thu Apr 23 08:36:46 UTC 2020 x86_64 GNU/Linux
```

**Additional context**

curiousleo:
Fixes #398.

## Overview

See https://github.com/target/lorri/issues/398#issuecomment-628464471 for context.

## Checklist

### DIFF
--- a/src/ops/start_user_shell.rs
+++ b/src/ops/start_user_shell.rs
@@ -60,18 +60,27 @@ PS1="(lorri) $PS1"
                 // https://superuser.com/a/591440/318156.
                 r#"
 unset RCS # disable automatic sourcing of startup scripts
+
+# reset ZDOTDIR
+if [ -z ${ZDOTDIR_BEFORE+x} ]; then
+    ZDOTDIR="${ZDOTDIR_BEFORE}"
+else
+    unset ZDOTDIR
+fi
+
+ZDOTDIR_OR_HOME="${ZDOTDIR:-${HOME}}"
 test -f "$ZDOTDIR_OR_HOME/.zshenv" && . "$ZDOTDIR_OR_HOME/.zshenv"
 test -f "/etc/zshrc"               && . "/etc/zshrc"
+ZDOTDIR_OR_HOME="${ZDOTDIR:-${HOME}}"
 test -f "$ZDOTDIR_OR_HOME/.zshrc"  && . "$ZDOTDIR_OR_HOME/.zshrc"
-PS1="(lorri) $PS1"
+
+PS1="(lorri) ${PS1}"
 "#,
             )
             .expect("failed to write zsh init script");
-            cmd.env(
-                "ZDOTDIR_OR_HOME",
-                env::var("ZDOTDIR")
-                    .unwrap_or_else(|_| env::var("HOME").expect("$HOME must be set")),
-            );
+            if let Ok(d) = env::var("ZDOTDIR") {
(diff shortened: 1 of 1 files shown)


Profpatsch:
Weak +1 on this, I’d eventually rather we just set `IN_LORRI_SHELL` and have an easy path for users to add it to their prompts manually. Otherwise we are going to have to add more and more logic on our side, which is just going conflict with user shell configuration, leading to more workarounds.

curiousleo:
> Weak +1 on this, I’d eventually rather we just set `IN_LORRI_SHELL` and have an easy path for users to add it to their prompts manually. Otherwise we are going to have to add more and more logic on our side, which is just going conflict with user shell configuration, leading to more workarounds.

Agreed, having shell-specific stuff in lorri is a maintenance burden. For now I think this PR fixes an issue without making it more complicated to put another solution in place later on, so I'll go ahead an merge.

Profpatsch:
SGTM

PR #402, opened 2020-05-14
Cole h/ci GitHub actions test macos

Profpatsch:
Branch mentioned in https://github.com/target/lorri/pull/396#issuecomment-626755910, let’s see what happens with MacOS actions. cc @cole-h


cole-h:
Just FYI (in case you didn't notice), there is also a commit that removes all the Travis stuff from around the repo -- feel free to drop that commit if you want to keep Travis around for a while longer.

Profpatsch:
I’d like to verify that both versions have the same featureset (or the
Actions pipeline is a superset) before I drop the travis file. But it’s
certainly looking very good.

Profpatsch:
I’ll close this for now, we have verified that MacOS builds seem to work without further setup.

PR #405, opened 2020-05-15
Simplify logged-evaluation.nix

curiousleo:
## Overview

`lorri services` is not being worked on anymore, but preliminary work towards it complicated `logged-evaluation.nix` significantly. This PR reverts those changes. There is more cleanup work to do, see #404.

## Checklist

No user-visible changes.

- [ ] Updated the documentation (code documentation, command help, ...)
- [ ] Tested the change (unit or integration tests)
- [ ] Amended the changelog in `release.nix` (see `release.nix` for instructions)

Issue #404, opened 2020-05-15
Remove NixFile::Services


curiousleo:
We're not working on `lorri services` anymore, but there are remnants of the beginnings of an implementation of it in the codebase.

Your mission, should you choose to accept it, will be to remove those remnants. Start with `NixFile::Services`, which is an enum value that is never produced in production, and go from there.

Issue #108, opened 2019-06-27
Services


grahamc:
@Profpatsch @shlevy and I had a meeting last week about managing development time services for lorri project environments.

# Background

The story here is for a developer whose software has run-time service dependencies, possibly complicated ones. These services should start when the user is working on the project and should eventually shut down after the user stops work.

Services might be easy to "pre-package", (for example, but not necessarily using NixOS modules): mysql, postgres, kafka, zookeeper

Or might be custom to the user's organization, like another repo might be used to build software, and run that software as a service.

Or it might even be provided by the project itself, where the repo builds several executables, and lorri starts/restarts them.

# Immediate Questions

How do we want to express service definitions? We should probably aim for _simple to configure_ than _very flexible_ .

Because a project might coexist on the same host as many other projects, do we want to ensure ports are _assigned_ instead of _requested_? 

Do we want to be able to run services _remotely_ like using `docker run`?

Do we want to be able to run services _in CI_?
# Long Term Implementation Ideas

 - "Compiling" service definitions to launchd or systemd units, for use in the "user" agents. Requires expertise in how launchd works for implementation and users (for debugging and diagnosing issues.)
 - Expressing the services using a subset of systemd's service definition format, and implementing a systemd-to-launchd converter (same trouble as above.)
 - Use something like s6 (possible other inits?) to manage process starting. Ideally something with minimal bash scripts necessary.
 - Build and launch services with Docker. Might work real nicely overall, though for services provided by the local repo, might be a bit awkward / slow.
 - Support many copies of a given service. Ideally we can tightly limit the access a service has to the host, so their state can be carefully isolated, and erased.
- Developers should be able to replace a local service with a remote service (ie: staging, production.) 

# Example application

The implementation of this feature should include an example application. Right now the thinking is a trivial "todo" manager written in Rust, backed by Redis.

Entering the directory and running `lorri services start` should start the services and make the webserver available on some local port.

Recompiling the code to the Rust application should cause the server to stop and start once the project has rebuilt. Because the application is hosted in the local repository, this restart should be triggered after the user runs `cargo build`.

# Dogfood

Developers working on this project should use lorri services to manage their project's dependencies, and report bugs or fixes for UX problems. I'll be using it on OfBorg, as it depends on several external services, and several services built out of the local source.

jurpeedurp:
[...]

Issue #104, opened 2019-06-21
lorri services: prior art link collection 


Profpatsch:
Let’s dump <s>interesting</s>relevant projects here for evaluation.

- https://kubernetes.io
- https://github.com/txn2/kubefwd
- https://kubernetes.io/docs/setup/learning-environment/minikube/
- https://microk8s.io/
- https://docs.docker.com/compose/

jurpeedurp:
https://skaffold.dev/docs/
https://draft.sh/

jurpeedurp:
https://www.nomadproject.io/

yurrriq:
I'd be curious to see a brief write-up of how the projects listed so far are relevant to lorri. Naively they seem unrelated, but maybe that's just based on my usage.

Separately, thanks for the passive recommendation of kubefwd :)

Profpatsch:
> I'd be curious to see a brief write-up of how the projects listed so far are relevant to lorri. Naively they seem unrelated, but maybe that's just based on my usage.

Having a fast & deterministic setup of a project’s environment is one thing, but developers often need running services, e.g. a local postgres with a test db or some microservices that one depends on (you run the services you are not actively developing, while iterating over the code).

We want to provide support for that use-case.

Profpatsch:
By @jurpeedurp, over at https://github.com/target/lorri/issues/108#issuecomment-506921801:
> I like the idea of using something like https://www.nomadproject.io/ under the hood because the wheel wouldn't need to be reinvented and it seems to serve many of the short and long term goals stated while still being flexible enough to meet future use cases.

> The following drivers are most likely of interest:
> https://www.nomadproject.io/docs/drivers/exec.html
> https://www.nomadproject.io/docs/drivers/docker.html

> Jobs can be provided in JSON / HCL. See the spec here:
> https://www.nomadproject.io/docs/job-specification/index.html

hanshoglund:
Also mentioned earlier by @grahamc 
> systemd's service definition format

yurrriq:
Thanks for the explanation, @Profpatsch. Sounds great!

nlewo:
https://github.com/rancher/k3s

manveru:
adding https://github.com/hercules-ci/arion since that combines nix and docker-compose :)

spinus:
I think arion would be good choice, already nix integrated into the stack, so only service restart needs to be implemented.

PR #407, opened 2020-05-18
Stabilize event stream

nyarly:
#Overview

At long last, stabilizes the event stream Varlink interface. Per discussion, this formalizes most of the varlink protocol for Lorri as private, exposing only the event stream interfaces. Events are set up as closer to a proper sum type, such that each kind has a separate sub-kind to describe it.

A follow up PR would move event-stream out from under the `internal` command and thus publish it.

## Checklist


build.rs
```diff
@@ -47,4 +47,10 @@ pub const RUN_TIME_CLOSURE: &str = "{runtime_closure}";
         "src/com.target.lorri.varlink",
         /* rustfmt */ true,
     );
+
+    // Generate src/com_target_lorri.rs
```
curiousleo, line 51:
```suggestion
    // Generate src/com_target_lorri_internal.rs
```

I think this calls for a `for_each` to process the two varlink files together, i.e. conceptually:

```
["src/com.target.lorri.varlink", "src/com.target.lorri.internal.varlink"]
    .for_each(|v| varlink_generator::cargo_build_tosource(v, /* rustfmt */ true);
```

release.nix
```diff
@@ -6,6 +6,11 @@
     # Find the current version number with `git log --pretty=%h | wc -l`
     entries = [
       {
+        version = 570;
+        changes = ''
+          Made the event stream Varlink events public, with sum-style types.
```
curiousleo, line 11:
```suggestion
          Make the event stream Varlink events public, with sum-style types.
```

For consistency - the release notes are written in the present tense.

src/com.target.lorri.internal.varlink
```diff
@@ -0,0 +1,72 @@
...
+)
+
+# Command describes how to run a terminal application.
+type Command (
+  # The path of the command binary.
+  program: string,
+
+  # Arguments to be passed to the binary.
+  args: []string
+)
```
curiousleo, line 72:
```suggestion
```

Services aren't a thing anymore - there is no need to copy this over. See https://github.com/target/lorri/issues/404 and https://github.com/target/lorri/pull/405.

src/daemon/rpc.rs
```diff
@@ -104,107 +112,108 @@ impl rpc::VarlinkInterface for Server {
...
-                result: re.result.ok_or("missing result!")?.into(),
-            },
-            failure => Event::Failure {
-                nix_file: re.nix_file.ok_or("missing nix file!")?.try_into()?,
-                failure: re.failure.ok_or("missing failure log")?.try_into()?,
-            },
+            section_end => build_loop::Event::SectionEnd,
+            started => re.reason.ok_or("missing reason!")?.try_into()?,
+            completed => re.result.ok_or("missing result!")?.try_into()?,
+            failure => re.failure.ok_or("missing failure log")?.try_into()?,
```
curiousleo, line 180: missing exclamation mark? :)

curiousleo: Actually, I don't think the exclamation marks are necessary. It's already in an error message. No need to draw more attention to it IMO.

src/daemon/rpc.rs
```diff
@@ -281,75 +296,131 @@ impl From<rpc::Outcome> for build_loop::BuildResults {
...
+                        msg: msg.clone(),
+                    }),
+                    exit: None,
+                    output: None,
+                    /*
+                    cmd: Some(cmd.into()),
+                    msg: Some(msg.into()),
+                    logs: None,
+                    status: None,
+                    */
```
curiousleo, line 331:
```suggestion
```

I guess this is no longer needed?

src/daemon/rpc.rs
```diff
@@ -281,75 +296,131 @@ impl From<rpc::Outcome> for build_loop::BuildResults {
...
+                        status: status.map(|i| i as i64),
+                        logs: logs.iter().map(|l| l.to_string()).collect(),
+                    }),
+                    output: None,
+                    /*
+                    cmd: Some(cmd.into()),
+                    logs: Some(logs.iter().map(|line| line.to_string()).collect()),
+                    msg: None,
+                    status: status.map(i64::from),
+                    */
```
curiousleo, line 349:
```suggestion
```

No longer needed?

src/daemon/rpc.rs
```diff
@@ -281,75 +296,131 @@ impl From<rpc::Outcome> for build_loop::BuildResults {
...
+                                                           .ok_or("exit error missing logs!")?
+                                                           .into_iter()
+                                                           .map(|l| l.into())
+                                                           .collect(),
+                                                           status: pf.status.map(|c| c as i32),
+                                                           },
+                                                           output => BuildError::Output {
+                                                           msg: pf.msg.ok_or("output failure without msg!")?,
+                                                           },
+                                                           */
```
curiousleo, line 422: No longer needed either?

src/lib.rs
```diff
@@ -31,11 +31,20 @@ pub mod watch;
...
-#[allow(missing_docs, clippy::all)]
+#[allow(missing_docs, clippy::all, unused_imports)]
 mod com_target_lorri;
 
+#[allow(missing_docs, clippy::all, unused_imports)]
+mod com_target_lorri_internal;
+
 #[allow(missing_docs)]
-pub mod rpc {
+pub(crate) mod rpc {
```
curiousleo, line 41: Perhaps a better name for this would now be `proto_internal`?

src/com.target.lorri.internal.varlink
nyarly: Skeptic though I was, I find that kind of a shame. I wonder if a separate tool might be worth examining? Or if something based on runit/s6 might be added to the pipeline?

nyarly:
Thanks for the review. Bunch of trivial oversights - the hazards of working in spare moments.

src/com.target.lorri.internal.varlink
curiousleo: If you have ideas, feel free to join the discussion at https://github.com/target/lorri/issues/104 and https://github.com/target/lorri/issues/108 :)

However, AFAICT nobody is working on this currently.

src/com.target.lorri.varlink
```diff
@@ -1,85 +1,58 @@
...
-# re-evaluates it as appropriate and sends a reply for each Command again.
-#
-# This is a streaming RPC. The daemon only accepts client calls with the "more"
-# property set - see https://varlink.org/Method-Call.
-# TODO: Implement WatchServices
-#method WatchServices(services_nix: ServicesNix) -> (service: Service)
-
-# ServicesNix describes the Nix expression which evaluates to a list of
-# services.
-type ServicesNix (
```
curiousleo, line 60: I used to design gRPC APIs a bunch. In doing so, I found the [Google API Design Guidelines](https://cloud.google.com/apis/design) a real help.

The style I settled on here is a combination of Google API Design Guidelines and the [Golang comment format](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences).

Concretely, this means:
- every type has a comment
- every not completely trivial field has a comment
- the comment format for type or field `X` is `X <verb> <rest of at least one full sentence>.`
- type and field names are full English words unless they are very long and the abbreviation is very common, e.g. `api` instead of `application_programming_interface`

Particularly if this is to be the project's public API, I think it is worth making the declarations in this file look consistent and documenting it.

src/ops/ping.rs
```diff
@@ -1,18 +1,21 @@
...
 use std::convert::TryFrom;
 
 /// See the documentation for lorri::cli::Command::Ping_ for details.
-pub fn main(nix_file: NixFile) -> OpResult {
-    // TODO: set up socket path, make it settable by the user
-    let address = crate::ops::get_paths()?.daemon_socket_address();
-    let shell_nix = rpc::ShellNix::try_from(&nix_file).unwrap();
+pub fn main(nix_file: NixFile, addr: Option<String>) -> OpResult {
+    let address = addr
+        .ok_or(Err(()))
```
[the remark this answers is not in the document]
nyarly: The chained `or_else` retains the existing error that can arise if if `get_paths` fails.

To be honest, this was what I got working to align the Option of a provided socket address (which I needed for the integration tests) to work with the Result produced by `get_paths`. 

It's possible that a `map_err` would produce a better result, but that would be the case across the board - alternatively, the `get_paths` Err _ought_ to be human readable, no?

So I guess:

0. Do you have a better solution for welding an Option and a Result together?
0. Shouldn't the `Err` values here be readable? If not, I think that's a separate Issue.
0. On reflection, this seems like the model for handling provided socket addresses in the first place, so it would be good to get it right here.

src/com.target.lorri.varlink
nyarly: The only quibble I'd make here is that the Go comment format (i.e. `X <verb> <sentence>`) arises in part because of the anemic feature set of godoc, and often leads to awkward and boring documentation. (Often I see `Read implements io.Reader on MyType`)

I'm currently adding documentation comments based on the rest of that guide, using the dictionary style that the Google guidelines examples use.

src/ops/ping.rs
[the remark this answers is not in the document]
curiousleo: My bad. The `Err(())` here is immediately dealt with on the next line. I was worried it could be "shown" to the user, containing no information. But that is not the case. I couldn't find a better way to express this as a chain, either. As big a fan of chaining as I am, the following strikes me as a more obvious way of doing the same thing:

```rust
    let address = match addr {
        Some(address) => address,
        None => crate::ops::get_paths()?.daemon_socket_address(),
    };
```

src/com.target.lorri.varlink
curiousleo: > I'm currently adding documentation comments based on the rest of that guide, using the dictionary style that the Google guidelines examples use.

Awesome, thank you!

Profpatsch:
Huh, this is certainly quite a diff. I will take a look soon, but I can’t promise I can finish a review in the next few days.

I usually expect an initial review of 500+ lines to take a block of 2–3 hours, which I can seldomly reserve for lorri in one go. It certainly helps if there are multiple commits that each do a small incremental part of the changes.

nyarly:
@Profpatsch What's your availability to review this looking like?

src/com.target.lorri.varlink
```diff
@@ -1,85 +1,104 @@
...
-  # The path of the command binary.
-  program: string,
+# Describes a failed build process
+type ExitFail (
+    # The command executed by Lorri
+    command: string,
+    # The Unix exit status of the command, if available
+    status: ?int,
+    # Log output of the command
+    logs: []string
```
Profpatsch, line 96: We should probably distinguish between stdout and stderr here, right? Either two lists, or one list with entries that have a tag whether they are stdout or stderr, which would keep their ordering.

src/com.target.lorri.varlink
```diff
@@ -1,85 +1,104 @@
...
+)
 
-  # Arguments to be passed to the binary.
-  args: []string
+# Describes a failure caused by output produced by the build that Lorri cannot
+# parse
+type OutputFail (
+    # A message describing the failure
+    message: string
```
Profpatsch, line 103: Really like the exhaustive documentation of all the fields.

src/lib.rs
```diff
@@ -65,6 +78,18 @@ impl From<&NixFile> for PathBuf {
     }
 }
 
+impl ToString for NixFile {
+    fn to_string(&self) -> String {
+        PathBuf::from(self).as_path().to_string_lossy().to_string()
```
Profpatsch, line 83: I don’t think we can have this instance, where is it used? Filenames are `[u8]` on unix, we shouldn’t convert like this.

src/lib.rs
```diff
@@ -65,6 +78,18 @@ impl From<&NixFile> for PathBuf {
...
+    }
+}
+
+impl From<String> for NixFile {
+    fn from(string: String) -> Self {
+        NixFile::Shell(PathBuf::from(string))
+    }
+}
```
Profpatsch, line 91: Where do we use this instance? It’s valid in theory, but going from a string to a PathBuf sounds like a design smell.

src/ops/ping.rs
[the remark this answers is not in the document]
Profpatsch: Yeah, let’s prefer `match` statements whenever possible, they are a lot easier to read usually.

src/daemon/internal_proto.rs
```diff
@@ -0,0 +1,433 @@
...
+impl TryFrom<&NixFile> for internal_proto::ShellNix {
+    type Error = &'static str;
+
+    fn try_from(nix_file: &NixFile) -> Result<Self, Self::Error> {
+        match nix_file {
+            NixFile::Shell(shell) => match shell.as_os_str().to_str() {
+                Some(s) => Ok(internal_proto::ShellNix {
+                    path: s.to_string(),
+                }),
+                None => Err("nix file path is not UTF-8 clean"),
```
Profpatsch, line 422: This looks like a bug to me, we shouldn’t require paths to be utf-8 clean, ever. Can varlink not transfer `[u8]`?

tests/daemon/main.rs
```diff
@@ -36,14 +36,9 @@ pub fn start_job_with_ping() -> std::io::Result<()> {
...
-    use lorri::rpc::VarlinkClientInterface;
-    rpc::VarlinkClient::new(connect(&address, Duration::from_millis(1000)))
-        .watch_shell(rpc::ShellNix {
-            path: shell_nix.to_str().unwrap().to_string(),
-        })
-        .call()
-        .unwrap();
+    connect(&address, Duration::from_millis(1000));
+
+    ping::main(lorri::NixFile::Shell(shell_nix), Some(address)).unwrap();
```
Profpatsch, line 41: nit
```suggestion
    lorri::ops::ping::main(lorri::NixFile::Shell(shell_nix), Some(address)).unwrap();
```

src/com.target.lorri.varlink
nyarly: :)

src/cli.rs
```diff
@@ -171,6 +171,9 @@ pub struct Ping_ {
     /// The .nix file to watch and build on changes.
     #[structopt(parse(from_os_str))]
     pub nix_file: PathBuf,
+    /// A custom socket address to ping - used mostly for testing and experiments
+    #[structopt(long = "socket-address")]
+    pub socket_address: Option<String>,
```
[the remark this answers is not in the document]
nyarly: If the socket address is changeable, it should be updated everywhere the socket is used. I've created #445 to track that.

I needed to add this in order to make integration tests run successfully, so the flag is really only used there. Does `structopt` provide a way to hide flags?

src/com.target.lorri.varlink
```diff
@@ -1,85 +1,104 @@
...
-  path: string
-)
-
-# Monitor the daemon. The method will reply with an update whenever a build begins or ends.
-# Montior will immediately reply with a snapshot of known projects, then a marker event,
-# indicating that the stream of events is now "live."
+# Monitor the daemon. The method will reply with an Event update whenever a
+# build begins or ends.  Monitor will immediately reply with a snapshot of
+# known projects, then a marker event, indicating that the stream of events is
+# now "live."
```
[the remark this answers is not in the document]
nyarly: I'll add to the docs here.

src/daemon/internal_proto.rs
nyarly: "Can it not" is unclear. 

Lorri uses a `PathBuf`, but varlink needs to ship a `String` (because JSON). What I found to do that was `shell.as_os_str().to_str()` which returns `None` if there's unclean bytes. In proper operation, that `None` branch should never get exercised, I think? When would `OsStr` not be representable as `String`?

src/lib.rs
nyarly: src/daemon/internal_proto.rs#320 (also L331 and 343)

nyarly: How should we convert otherwise?

src/lib.rs
nyarly: This is more generally used in `internal_proto` - e.g. L246, where it's used to push the path into a `Reason::ProjectAdded`

src/ops/ping.rs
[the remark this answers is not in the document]
nyarly: Updating this to use match was much better.

nyarly:
@Profpatsch Freshly merged to master.

Profpatsch:
I squashed the commits so far and gave them slightly better commit messages.

Looking over the code, there are some places where strange conversion instances are implemented (e.g. a `ToString` instance for `NixFile`, I will need to look through them one by one before merging).

Profpatsch:
Thinking about filename fields

> How can I reference a file?
>
> Varlink should be remotable, therefore you should not point a remote service to a file, which is local to the client. Just send the content and an identifier/name of the file.

Though varlink has no bytestring scalar, so I have no idea how they intend us to send the content of files?
I’m really not sure we want to stick with varlink as the interface protocol. It feels super underspecified and not very well thought-through to me.

For now I guess the best we can do is manually decode to a unicode string and fail if a filename contains non-utf8 bytes (which is horrible haha). There is just no way to encode arbitrary bytes via varlink (other than going via the base64 hack).

What is really not a good idea is using the lossy conversion, because the whole reason we are exposing the interface is that we want to enable external clients to get the data (and lossy conversion does not give us the data).

src/com.target.lorri.varlink
Profpatsch: It’s only stderr, I changed the docs to make that clear.

Issue #406, opened 2020-05-15
Ensure nixos-19.09 compatibility in CI


curiousleo:
**Describe the bug**

We would like lorri to be compatible with NixOS 19.09, but our CI only builds and tests against 20.03.

More context here: https://github.com/target/lorri/pull/396#discussion_r422865040.

**To Reproduce**

1. Go to https://github.com/target/lorri/commit/d3ab9206f95c88ac692324cf8554477d28337fb6#diff-4b7dc34deb385d784093e05de88b9078R7
2. See that it says "nixos-20.03"
3. ...

**Expected behavior**

If we want to [keep supporting NixOS 19.09](https://github.com/target/lorri/pull/396#discussion_r425230375), we need to build and test lorri against 19.09 in CI.

cole-h:
Should this just be a revert of my commit bumping to 20.03 (sorry), or would using Rust 1.37.0 from a pinned nixpkgs-mozilla overlay be sufficient to ensure 19.09 support?

If we want to use 19.09 for everything, I'd like to suggest also having 20.03 around, if only to get a cached `nixpkgs-fmt` that doesn't need to be rebuilt in CI (although, I guess cachix support for CI would mostly alleviate this issue after the first run).

curiousleo:
> Should this just be a revert of my commit bumping to 20.03 (sorry), or would using Rust 1.37.0 from a pinned nixpkgs-mozilla overlay be sufficient to ensure 19.09 support?

First of all, no worries! This is what code review is there for in my opinion: to give context - it's not obvious that 19.09 was a conscious choice of channel to build against just from the build setup.

Secondly, I think that the most obvious way to ensure compatibility with 19.09 is to build and test against 19.09. I think that building against Rust 1.37.0 with `nixpkgs-mozilla` overlay and other packages coming from 20.03 makes it _very likely_ that lorri is 19.09-compatible, but there's no guarantee.

> If we want to use 19.09 for everything, I'd like to suggest also having 20.03 around, if only to get a cached `nixpkgs-fmt` that doesn't need to be rebuilt in CI (although, I guess cachix support for CI would mostly alleviate this issue after the first run).

[...]

PR #411, opened 2020-05-29
ci: ensure compatibility with 19.09

cole-h:
There are a few steps I've taken to make this as robust as possible:

0) Add a `nixpkgs` arg to default.nix and shell.nix, allowing us to more
easily specify which nixpkgs to use -- rather than having to use
`nix-build --arg pkgs 'import ./nix/nixpkgs-1909.nix'`, one only needs
to run `nix-build --arg nixpkgs ./nix/nixpkgs-1909.nix`.

1) Create two separate nixpkgs expressions, nixpkgs-stable and
nixpkgs-1909. This allows us to build against NixOS 19.09 by running the
aforementioned `nix-build --arg nixpkgs ./nix/nixpkgs-1909.nix`
commandline. (Also add nixos-19.09 to the update-dependencies script.)

2) Add an `overlay` build step that overlays the lorri available from
the current stable nixpkgs, to ensure that it is still possible to do
so. ("in order to not have to wait on a new lorri release in nixpkgs" --
Profpatsch)

---

Fixes https://github.com/target/lorri/issues/406.

Also, I updated to the latest action because it lets us drop the use of NIX_PATH/channels since we use our own versions stored in the `.json`s (which ever-so-slightly speeds up CI).

Let me know if there's anything you would like to see improved.

cc @curiousleo @Profpatsch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,5 +81,7 @@ jobs:
         uses: cachix/install-nix-action@v9
         with:
             skip_adding_nixpkgs_channel: true
-      - name: Build w/ overlay
-        run: nix-build ./nix/overlay.nix -A lorri
+      - name: Build w/ overlay (19.09)
+        run: nix-build ./nix/overlay.nix -A lorri --arg pkgs ./nix/nixpkgs-1909.json
+      - name: Build w/ overlay (stable)
+        run: nix-build ./nix/overlay.nix -A lorri --arg pkgs ./nix/nixpkgs-stable.json
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,15 @@ on:
   pull_request:
     branches: [ "**" ]
 jobs:
-  build:
+  rust:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Nix
-        uses: cachix/install-nix-action@v7
+        uses: cachix/install-nix-action@v9
+        with:
+            skip_adding_nixpkgs_channel: true
       - name: Cache cargo registry
         uses: actions/cache@v1
         with:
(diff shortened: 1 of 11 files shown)


nix/overlay.nix
```diff
@@ -0,0 +1,14 @@
+let
+  srcDef = builtins.fromJSON (builtins.readFile ./nixpkgs-stable.json);
```
Profpatsch, line 2: I think the idea behind this was to also check current lorri against nixpkgs 19.09, so shouldn’t there be a second overlay for testing current master of this repository, building it against 19.09? I might be missing the point.

cole-h: Oops. You are indeed correct.

Do you want two overlays-tests, or just the one (for 19.09)?

Profpatsch:
Hm, before we just built lorri with the pinned nixpkgs via default.nix
iirc, so why not just have it build the `default.nix` twice, once with
19.09 and once with 20.03?

.github/workflows/ci.yml
```diff
@@ -37,21 +39,49 @@ jobs:
...
+  overlay:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Nix
+        uses: cachix/install-nix-action@v9
+        with:
+            skip_adding_nixpkgs_channel: true
+      - name: Build w/ overlay (19.09)
```
cole-h, line 84: @Profpatsch I'm not sure I understood you correctly, but is this (similar to) what you had in mind?

curiousleo: @Profpatsch ^

nix/overlay.nix
curiousleo: @Profpatsch ^

Profpatsch:
Perfect, thank you.

PR #421, opened 2020-06-16
trivial-shell.nix: use with

asymmetric:
To clean up list of packages.

I'm assuming this should not be mentioned in the release notes, since it's a visible and backwards-compatible change with no effect on lorri's semantics, but let me know if it should instead.

## Checklist

### DIFF
--- a/src/trivial-shell.nix
+++ b/src/trivial-shell.nix
@@ -1,10 +1,10 @@
 { pkgs ? import <nixpkgs> {} }:
 
 pkgs.mkShell {
-  buildInputs = [
-    pkgs.hello
+  buildInputs = with pkgs; [
+    hello
 
     # keep this line if you use bash
-    pkgs.bashInteractive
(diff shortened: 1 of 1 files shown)


asymmetric:
I agree with you @curiousleo, `with` can easily be abused. But I think that used this way, on the RHS of an assignment, limits its nastier aspects. :)

curiousleo:
@Profpatsch @grahamc @nyarly a second reviewer's opinion perhaps?

Profpatsch:
yeah, sorry, I’m gonna -1 this one, I manually prefixed `pkgs` as a deliberate choice, because I don’t think a file like this should contain an advanced language feature (and frankly an anti-pattern) like `with`.

Profpatsch:
But thank you @asymmetric for your PR, we have a bunch of issues marked as good first issues if you want to take a shot.

PR #430, opened 2020-06-23
Run ci_check instead of individual `cargo` commands

curiousleo:
Brings GitHub actions CI closer to what Travis does right now.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,14 +36,8 @@ jobs:
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
       - name: Shell (cache inputs)
         run: nix-shell
-      - name: Format
-        run: nix-shell --run 'cargo fmt --all -- --check' --arg isDevelopmentShell false
-      - name: Build
-        run: nix-shell --run 'cargo build' --arg isDevelopmentShell false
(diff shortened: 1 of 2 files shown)


curiousleo:
@Profpatsch any idea what is going on here? https://github.com/target/lorri/pull/430/checks?check_run_id=800250471#step:7:7

If that doesn't ring a bell, I'll investigate tomorrow.

Profpatsch:
Some `?` blows up.

We should add a flag that makes lorri go into “don’t be user-friendly” mode and just print the crash log.

.github/workflows/ci.yml
```diff
@@ -60,6 +54,10 @@ jobs:
             signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
       - name: Build
         run: nix-build
+      - name: Install
+        run: nix-env -i ./result
+      - name: Test self-upgrade (TODO)
+        run: echo "lorri self-upgrade local $(pwd)"
```
curiousleo, line 60: `lorri self-upgrade local $(pwd)` blows up right now, I'll deal with that in a separate PR.

curiousleo: See #432.

curiousleo:
@Profpatsch tests passed -- is this good to go?

Issue #375, opened 2020-04-13
Failed to join thread, despite catch_unwind!: Any


Savir:
**Describe the bug**

**To Reproduce**

1. started lorry as a user daemon (`systemctl --user restart lorri`)
2. logged into a directory with a `default.nix` config file
3. check the logs `journalctl --user -fu lorri    `

**Expected behavior**

**Metadata**

```
Apr 13 13:58:51 hector-tests systemd[1579]: lorri.service: Main process exited, code=exited, status=101/n/a
Apr 13 13:58:51 hector-tests systemd[1579]: lorri.service: Failed with result 'exit-code'.
Apr 13 13:58:51 hector-tests systemd[1579]: lorri.service: Service hold-off time over, scheduling restart.
Apr 13 13:58:51 hector-tests systemd[1579]: lorri.service: Scheduled restart job, restart counter is at 1.
Apr 13 13:58:51 hector-tests systemd[1579]: Stopped Lorri Daemon.
Apr 13 13:58:51 hector-tests systemd[1579]: Started Lorri Daemon.
Apr 13 13:58:51 hector-tests lorri.sh[10958]: Apr 13 13:58:51.268 INFO ready
Apr 13 14:02:05 hector-tests lorri.sh[10958]: Apr 13 14:02:05.409 INFO build status, message: BuildEvent(Started { nix_file: Shell("/home/hector/Documents/tests/Projects/test/shell.nix"), reason: ProjectAdded(Shell("/home/hector/Documents/tests/Projects/test/shell.nix")) })                                                                                                       
Apr 13 14:03:44 hector-tests lorri.sh[10958]: Well, this is embarrassing.
Apr 13 14:03:44 hector-tests lorri.sh[10958]: lorri had a problem and crashed. To help us diagnose the problem you can send us a crash report.
Apr 13 14:03:44 hector-tests lorri.sh[10958]: We have generated a report file at "/tmp/report-99e37834-4263-41c8-85d5-180afe58cc98.toml". Submit an issue or email with the subject of "lorri Crash Report" and include the report as an attachment.
Apr 13 14:03:44 hector-tests lorri.sh[10958]: - Homepage: https://github.com/target/lorri
Apr 13 14:03:44 hector-tests lorri.sh[10958]: - Authors: Graham Christensen <graham.christensen@target.com>
Apr 13 14:03:44 hector-tests lorri.sh[10958]: We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.
[...]

PR #435, opened 2020-06-25
Thread.rs correctly unwind call stacks

Profpatsch:
This is a rather bigger change than I hoped it would be, but I think I found a good solution for the issues many people have been having in https://github.com/target/lorri/issues/375.

Which is mainly that the thread pool was not re-throwing the call stack of the panicking inner thread, but just throwing its own panic, so `human-panic` would show the same error message for every panic (that the `dyn Any` returned by `thread_handle.join()` cannot print the actual data did not help).

I did a bit of refactoring before tackling the actual problem and wrote quite extensive commit messages, so best reviewed commit-by-commit.

I started using [commit markers](https://code.tvl.fyi/tree/docs/CONTRIBUTING.md#n35) to better distinguish refactor commits from actual feature commits, and it looks quite usable to me.

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -8,33 +8,32 @@
 //! already.
 
 use crossbeam_channel as chan;
+use std::any::Any;
 use std::collections::HashMap;
 use std::thread;
 use std::thread::ThreadId;
-
-struct DeathCertificate {
-    tx: chan::Sender<ThreadId>,
-}
-
-impl Drop for DeathCertificate {
-    fn drop(&mut self) {
-        self.tx
-            .send(thread::current().id())
-            .expect("failed to send thread shut-down message!");
-    }
-}
 
 struct Thread {
     name: String,
     join_handle: thread::JoinHandle<()>,
 }
 
+struct Dead {
+    thread_id: ThreadId,
+    cause: Cause,
+}
+
+enum Cause {
+    Natural,
+    Paniced(Box<dyn Any + Send>),
+}
+
 /// A thread pool for joining many threads at once, panicking
 /// if any of the threads panicked.
 pub struct Pool {
     threads: HashMap<ThreadId, Thread>,
-    tx: chan::Sender<ThreadId>,
-    rx: chan::Receiver<ThreadId>,
+    tx: chan::Sender<Dead>,
+    rx: chan::Receiver<Dead>,
 }
 
 impl Default for Pool {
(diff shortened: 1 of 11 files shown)


src/thread.rs
```diff
@@ -63,27 +62,38 @@ impl Pool {
...
+                    .send(Dead {
+                        thread_id,
+                        cause: Cause::Paniced(panic),
+                    })
+                    .expect("failed to send thread shut-down message!"),
+            }
+            tx.send(Dead {
+                thread_id,
+                cause: Cause::Natural,
+            })
```
Profpatsch, line 89: This must be in the `Ok` above, or two messages are sent

Profpatsch: I changed it to return only the cause in the match itself and send one message immediately after.

Profpatsch:
Adding 

```
        pool.spawn("nooo", || {
            panic!("in the disco");
        })?;
```

In between the spawns in `daemon.rs` and running `lorri daemon` gives:

```
> lorri daemon
Jun 30 00:41:24.900 INFO ready
Well, this is embarrassing.

lorri had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/tmp/report-40262958-b316-4b0f-8d1d-ca3a15c5e10a.toml". Submit an issue or email with the subject of "lorri Crash Report" and include the report as an attachment.

- Homepage: https://github.com/target/lorri
- Authors: Graham Christensen <graham.christensen@target.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
philip@shiki ~ [101]> cat /tmp/report-40262958-b316-4b0f-8d1d-ca3a15c5e10a.toml
name = 'lorri'
operating_system = 'unix:Unknown'
crate_version = '1.1.0'
explanation = '''
Panic occurred in file 'src/daemon.rs' at line 109
'''
cause = 'in the disco'
method = 'Panic'
backtrace = '''

   0: 0x55563373312a - __rust_maybe_catch_panic
   1: 0x5556334f465c - std::sys_common::backtrace::__rust_begin_short_backtrace::h6433debb75883924
   2: 0x55563373312a - __rust_maybe_catch_panic
   3: 0x55563357a9e6 - core::ops::function::FnOnce::call_once{{vtable.shim}}::h6a8d7b46fd9b8ac7
   4: 0x555633726f0f - <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once::he9dcb7073be13c65
   5: 0x555633714a40 - std::sys_common::thread::start_thread::h9ce8f903e84fe8e9
   6: 0x555633726586 - std::sys::unix::thread::Thread::new::thread_start::h5e894f0575174d9e
   7: 0x7f4119516edd - start_thread
   8: 0x7f411942da4f - __clone
   9:        0x0 - <unknown>'''
```

So we get the good stack traces now.

PR #438, opened 2020-06-25
doc(MAINTAINERS.md): get rid of stable version-specific bits

Profpatsch:


MAINTAINERS.md
```diff
@@ -66,9 +66,9 @@ To cut a new release:
 
 ## Publishing a release on [nixpkgs][]
 
-lorri is available in the `nixos-unstable` and `nixos-19.09` release channels,
-which correspond to the `master` and `release-19.09` branches in the
-[nixpkgs][] repository, respectively.
+lorri is available in the `nixos-unstable` and the stable release channels,
```
curiousleo, line 69: "the stable release channels" would include 18.09. lorri is not currently in there. Are you saying that we should backport lorri to 18.09?

MAINTAINERS.md
```diff
@@ -86,9 +86,10 @@ To update the lorri version in [nixpkgs][]:
 
    > @GrahamcOfBorg build lorri.tests
 
-2. **`nixos-19.09`**: _after_ the first PR has been merged into `master`,
+2. **`nixos` stable**: _after_ the first PR has been merged into `master`,
    follow the [backporting procedure][nixpkgs-backporting] to create a PR
-   against `release-19.09`; see for example [NixOS#77432][nixos-stable-pr].
+   against `release-<stable-release-date>`;
+   see for example [NixOS#77432][nixos-stable-pr].
```
curiousleo, line 92: If we support multiple stable release channels, then these instructions need to be clearer IMO. Would the backporting need to happen in a particular order?

MAINTAINERS.md
Profpatsch: Ah, I didn’t want to say all stable channels of course. Rewording.

MAINTAINERS.md
Profpatsch: I would only ever backport to the latest stable and skip on major versions.

Profpatsch:
I pushed the fixup in a new commit and added a sentence concerning backports

MAINTAINERS.md
```diff
@@ -86,9 +86,11 @@ To update the lorri version in [nixpkgs][]:
 
    > @GrahamcOfBorg build lorri.tests
 
-2. **`nixos-19.09`**: _after_ the first PR has been merged into `master`,
+2. **latest `nixos` stable**: _after_ the first PR has been merged into `master`,
+   if the new release is not a major version bump (aka a breaking change),
    follow the [backporting procedure][nixpkgs-backporting] to create a PR
-   against `release-19.09`; see for example [NixOS#77432][nixos-stable-pr].
+   against `release-<latest-stable-release-date>` (e.g. `release-20.03`);
```
curiousleo, line 92: So we no longer update 19.09?

Profpatsch: Hm, we are halfway into 20.03. But maybe we should?

Profpatsch: I don’t know if the maintainers accept anything but a patch/security release at this point.

curiousleo: I have no opinion on this, but I would like the text to be clear on this point :)

Profpatsch: Hm, we do try to keep compatibility back with rustc as far as reasonable, but for nixpkgs stable releases I think we can target backports for the latest stable. Those two are not opposed to each other methinks, because you never know which situations people try to get lorri working in. 

e.g. I assume many people are stuck on way old nixpkgs pins in their company repos, so putting lorri in an overlay/flake would still work for them if we make sure we are compatible with older rustc. 

But for NixOS backports, people usually don’t expect any major bumps (and often not even minor bumps, because those could introduce new bugs if they add nontrivial features).

Profpatsch: Should I add that reasoning to the doc?

curiousleo: I think it's not necessary to include all the reasoning, but if I'm a new maintainer for lorri, I'd like to know what I need to do. So I just wanted to clarify that this was what you meant.

PR #436, opened 2020-06-25
chore(CI): remove travis CI

Profpatsch:
Thanks to the awesome work by @cole-h, we can remove Travis.

PR #442, opened 2020-06-30
chore(ci): run on darwin too

cole-h:
Followup to https://github.com/target/lorri/pull/436#issuecomment-651221145.

Next, I'd like to set up clippy and rustfmt to run on PRs. This would only happen on `ubuntu-latest`, because apparently, darwin can sometimes ship without `clippy`/`rustfmt` (see the failed CI on https://github.com/cole-h/lorri/pull/4 for examples of what happens when you try)?

cc @curiousleo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,10 @@ env:
   LORRI_NO_INSTALL_PANIC_HANDLER: absolutely
 jobs:
   rust:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


curiousleo:
@cole-h thank you!

> Next, I'd like to set up clippy and rustfmt to run on PRs.

Both are already running in CI via `ci_check` -- see https://github.com/target/lorri/blob/11693cfd38e7a8d1bf5c16c4e7f6c24c4f8b0479/.github/workflows/ci.yml#L45 and https://github.com/target/lorri/blob/11693cfd38e7a8d1bf5c16c4e7f6c24c4f8b0479/shell.nix#L151

cole-h:
Oh... Right. Thanks for reminding me :-D

PR #450, opened 2020-07-20
Add manpage

Profpatsch:
Habemus man page. See commits.

screenshot

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -56,13 +56,16 @@ To cut a new release:
    change the `version` field in `Cargo.toml` accordingly.
 2. Build the project to update `Cargo.lock`, then run `nix/update-carnix.sh` to
    update `Cargo.nix`.
-3. Create a PR with these changes and merge it. Note the hash of the merge
+3. Go through all commits since the last release and cross-check against the
+   release notes in `release.nix`. Add missing changes.
+4. With the changelog in mind, check that the manpage is up-to-date. 
+5. Create a PR with these changes and merge it. Note the hash of the merge
    commit.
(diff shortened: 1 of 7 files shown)


MAINTAINERS.md
```diff
@@ -54,15 +54,18 @@ does not in itself necessitate a major release.
 To cut a new release:
 1. Determine if this is a [minor or major release](#versioning-scheme) and
    change the `version` field in `Cargo.toml` accordingly.
-2. Build the project to update `Cargo.lock`, then run `nix/update-carnix.sh` to
+1. Build the project to update `Cargo.lock`, then run `nix/update-carnix.sh` to
```
curiousleo, line 57: Not sure about this. Yes, GitHub renders this as 1. 2. 3., but IMO Markdown is meant to be readable without any rendering at all. So I'm in favour of numbering lists 1. 2. 3. instead of 1. 1. 1.

Profpatsch: Sure, reverting.

Profpatsch:
>  I was also thinking that perhaps there are more human-readable / writable formats than mdoc for man pages?

I’m on the edge here.
I kinda used it as an excuse to learn how mdoc works, and I agree the readability is not great if you are not used to it.

However, the documentation of mdoc is quite fantastic, and it’s basically made for exactly this use-case. It can even be rendered by `man` as-is, no conversion toolchain required.

The same author also has the [`lowdown`](https://kristaps.bsd.lv/lowdown/) tool, which produces mdoc/html/… from markdown, but explicitly recommends ([see footnote](https://kristaps.bsd.lv/lowdown/#fn1) using plain mdoc for command manpages (and lowdown only for technical documents).

PR #451, opened 2020-07-21
release 1.1.1

Profpatsch:
I’d like to push the manpage change to nixpkgs.

The MAINTAINERS says we don’t use the patch version field, but I don’t think it’s wise to bump the minor version for an update in documentation, so I’d use the patch field in this case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "lorri"
 # lorri uses a MAJOR.MINOR versioning scheme, see MAINTAINERS.md for details.
 # Cargo requires MAJOR.MINOR.PATCH. So we translate lorri's versioning scheme
 # to MAJOR.MINOR.0, i.e. the PATCH part is always zero.
-version = "1.1.0" # Format: MAJOR.MINOR.0
+version = "1.1.1" # Format: MAJOR.MINOR.0
 authors = [
   "Graham Christensen <graham.christensen@target.com>",
 ]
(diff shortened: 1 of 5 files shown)


nyarly:
This LGTM. A man page would be very nice to have and have released.

curiousleo:
> The MAINTAINERS says we don’t use the patch version field, but I don’t think it’s wise to bump the minor version for an update in documentation, so I’d use the patch field in this case.

I see no harm in increasing the minor version. But if you want to do it like you said, go ahead. Before you do, please change MAINTAINERS accordingly to describe under which circumstances a change warrants a patch version increase. MAINTAINERS is useless if it doesn't reflect actual practice, and releasing a patch version currently contradicts what MAINTAINERS says.

Profpatsch:
I’m gonna ignore the clippy failure, it’s unrelated and will pop up in the next PR.

Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "lorri"
 # lorri uses a MAJOR.MINOR versioning scheme, see MAINTAINERS.md for details.
 # Cargo requires MAJOR.MINOR.PATCH. So we translate lorri's versioning scheme
 # to MAJOR.MINOR.0, i.e. the PATCH part is always zero.
```
curiousleo, line 5: Please update or remove this comment, it is now out of date.

PR #459, opened 2020-08-04
CI: dedup and trim

Profpatsch:
This revives the code we used for the travis config, which enabled us
to deduplicate the uncomfortably large amount of CI config. :)

It starts out by converting the config to nix, then refactoring until the gist of the jobs was readable again.

Then it removes some of the build steps, to bring down evaluation time.
Also updates the rust cache action.

cc @cole-h

### DIFF
--- a/.github/workflows/ci.nix
+++ b/.github/workflows/ci.nix
@@ -107,23 +107,6 @@ let
       };
     };
 
-    nix-shell = { runs-on }: {
-      name = "nix-shell-${runs-on}";
-      value = {
-        name = "nix-shell allBuildInputs (${runs-on})";
-        inherit runs-on;
-        steps = [
-          (checkout {})
-          setup-nix
-          setup-cachix
-          {
-            name = "Build";
-            run = "nix-build -A allBuildInputs shell.nix";
-          }
-        ];
-      };
-    };
-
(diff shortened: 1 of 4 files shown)


Profpatsch:
My opinion here is partially a principled one. As people who know how to solve the problems with the current technological choices, we should try to lead by example :)

Even if it comes at a slight compatibility cost.

lorri is a special case even in that space:

> The project is about experimenting with and improving the developer's
experience with Nix.

So part of the R&D contributed by this project is to serve as inspiration of how things *could* look like if the ideas of nix are applied to (dev)ops.

cole-h:
A bit late, but this is pretty cool!

A question: why does the `"with"` key need quotes? Seems a bit out of place, considering none of the other keys do this.

Profpatsch:
> A question: why does the "with" key need quotes? Seems a bit out of
place, considering none of the other keys do this.

`with` is a keyword in nix (e.g. `with import <nixpkgs> {}`), so it has to
be quoted. It does throw a syntax error if you forget, so it’s not a big
problem in practice.

cole-h:
Oh. Duh. As you can tell, it's been a little bit since I was in the Nix mindset :D

PR #471, opened 2020-08-25
feat(sigterm): daemon exits with 0 on SIGINT or SIGTERM

### LINKED ISSUE #463: Ctrc-C ing (SIGINT/SIGTERM) the daemon returns status 130/143 instead of 0
We should accept at least SIGTERM as valid shutdown request, if not SIGINT (which is useful on the command line).

On the command line it never saves `lorri daemon` in my history, because the return code is not successful.

tshaynik:
Make the `lorri daemon` subcommand exit with 0 on SIGINT
or SIGTERM.

Other `lorri` subcommands preserve the default SIGINT
and SIGTERM behaviors.

This change is done adds the [ctrlc](https://github.com/Detegr/rust-ctrlc)
crate, by Antti Keränen, as a dependency. This crate provides a simple
and safe interface for handling SIGINT and SIGTERM.

An alternative to the `ctrlc` crate would be to implement the response to the signals
using the the `nix` crate, which is already a lorri dependency.
The downside of this would be adding `unsafe` code to the codebase.

Closes #463 .

## Overview

* added the ctrlc crate as a dependency
* added a function that exits with 0 for SIGINT or SIGTERM
* call that function on the Daemon arm of the match statement in `run_command()`

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 [dependencies]
 atomicwrites = "0.2.3"
 crossbeam-channel = "0.3.8"
+ctrlc = { version = "3.1.6", features = ["termination"] }
 directories = "1.0.2"
 lazy_static = "1.2.0"
 md5  = "0.6.1"
--- a/src/main.rs
+++ b/src/main.rs
@@ -52,6 +52,14 @@ fn install_panic_handler() {
     }
 }
 
+// Exit with return code 0 on SIGINT and SIGTERM
+fn install_signal_handler() {
+    ctrlc::set_handler(move || {
+        std::process::exit(0);
+    })
+    .expect("Error setting SIGINT and SIGTERM handler");
+}
+
 /// Try to read `shell.nix` from the current working dir.
 fn get_shell_nix(shellfile: &PathBuf) -> Result<NixFile, ExitError> {
     // use shell.nix from cwd
@@ -105,6 +113,7 @@ fn run_command(log: slog::Logger, opts: Arguments) -> OpResult {
             watch::main(project, opts)
         }
         Command::Daemon(opts) => {
+            install_signal_handler();
             let _guard = without_project();
             daemon::main(opts)
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2018"
 
 [dependencies]
 atomicwrites = "0.2.3"
+ctrlc = { version = "3.1.6", features = ["termination"] }
 crossbeam-channel = "0.3.8"
```
curiousleo, line 17: "cr[ossbeam-channel]" comes before "ct[rlc]" in the alphabet, would you mind moving the new dependency down by one?

src/main.rs
```diff
@@ -52,6 +52,14 @@ fn install_panic_handler() {
     }
 }
 
+// Exit with return code 0 on SIGINT and SIGTERM
+fn install_signal_handler() {
+    ctrlc::set_handler(move || {
+        std::process::exit(0);
+    })
+    .expect("Error setting SIGINT and SIGTERM handler");
```
curiousleo, line 60: Given that all the signal handler itself does is provide a slightly nicer UX, crashing lorri if the handler cannot be set seems excessive. Under what circumstances would setting the handler fail?
- If we expect `set_handler` to fail every now and then, I would propose just logging a warning on failure and moving on.
- If `set_handler` fails "basically never", then calling `expect` here seems fine.

curiousleo (commented):
Great stuff! I have one nit and one question about `expect`. Thanks for your contribution @tshaynik!

Cargo.toml
tshaynik: Thanks! Just fixed this.

src/main.rs
tshaynik: Thanks. this makes sense as a concern!

Judging from the code and the variants of the  `ctrlc::Error` [enum](http://detegr.github.io/doc/ctrlc/enum.Error.html) it seems like it `set_handler` can fail  if:
1. The SIGINT or SIGTERM signals don't exist on the underlying system 
2. a programmer error where multiple Ctrl-C handlers have been registered
3. an expected system error (seems like it can only happen when basic Unix operations fail in surprising ways).

So it seems to me that `set_handler` should fail "basically never". 

I'm happy to change it to log a warning on failure if you prefer! But in that case, one concern might be that if it were to fail in that exceptional way and we don't exit the process immediately, `lorri daemon` might not respond to SIGINT or SIGTERM at all (I'm not sure how justified this concern is).

curiousleo: > So it seems to me that `set_handler` should fail "basically never".

Thanks for providing this context! Given your analysis, I agree - `expect` seems like the right thing to use here.

curiousleo (approved):
:1st_place_medal:

PR #478, opened 2020-10-01
doc(README): add FAQ entry for whether to check in .envrc

Profpatsch:
This is a frequently asked question, and we have an answer.

cc @MMesch who last asked the question.


README.md
```diff
@@ -238,6 +238,39 @@ By default, the upgrade command will upgrade from the `master` branch.
...
 
+## FAQ
+
+### Should I check the `.envrc` into git/my source control?
+
+Yes, you can!
+
+`direnv` will ask to run `direnv allow` every time the `.envrc` changes,
+and `lorri direnv` will never run anything else than what is required
+to set up the nix environment, so from a security standpoint it is safe.
```
curiousleo, line 249: > from a security standpoint it is safe

Assuming Nix expression cannot escape their sandbox, etc. I would not make security claims like this.

Suggestion: "it is _as safe as Nix itself_", as the security claim basically boils down to "it doesn't add much insecurity on top of Nix".

nyarly:
FTR I don't like this answer. The fact that Lorri has an .envrc checked in is an ongoing friction for me, because I use direnv for more than Lorri does. e.g. I use it to open or connect to a Tmux session automatically when I enter a project; my Lorri workspace behaves differently from every other project I work on as a result.

At a barest minimum, I'd think that the etiquette here is to have the project `.envrc` source something like `.envrc-local` so that each dev can use direnv as their used to.

README.md
Profpatsch: I don’t see how the restriction is super helpful here, because it makes people second-guess what that means. lorri does not any additional security risk (at least none that we now of) and the user is already using nix so they accepted the security risks that nix might have.

Profpatsch:
> At a barest minimum, I'd think that the etiquette here is to have the project `.envrc` source something like `.envrc-local` so that each dev can use direnv as their used to.

Good suggestion, I added a PR that adds that.

README.md
curiousleo: I would simply never publicly claim that something is secure without qualification if there hasn't at least been a third-party audit. It just too strong a claim to make without concrete, specific evidence that someone with relevant training has tried to circumvent the security measures you've put in place and did not succeed.

My suggestion would simply be to remove ", so from a security standpoint it is safe". The relevant bit for users is "will never run anything else than what is required".

Profpatsch: Good point, I removed that part of the sentence.

PR #483, opened 2020-11-12
Download systemd files with wget and respect XDG

xerus2000:
Only a usability improvement for the documentation


curiousleo:
Thanks for your contribution @xerus2000! I've taken the liberty to pick out the main change, namely downloading the systemd files so people don't have to check out the repo first - it's a really nice UX improvement!

PR #498, opened 2021-03-11
Only regenerate shell_gc_root if store path changes

### LINKED ISSUE #387: Do not bother user with rebuilds that don’t change the result
I _believe_ this is not a precise duplicate of any existing issues, but I might be wrong, in that case, sorry.

My `nix-shell` involes some evaluation-time computation that is easily triggered (in fact, it reads `.git/index`). Lorri detects that it uses that file (good), watches that file for changes (good), re-starts a build when that files changed (good), and when the build is finished, even thought the resulting derivation is identicaly, reloads my direnv (bad, because very noisy).

Maybe this is naive, but couldn’t `lorri` notice that the shell derivation is unchanged in most cases, and not do anything?

leotaku:
Currently, lorri may regenerate the shell_gc_root link to the project
environments store path, even if that store path has not changed.
With this commit, lorri will now check if the current and previous
store paths are identical and, if so, avoid regenerating the symlink.

This is especially interesting as lorri instructs direnv to watch the
shell_gc_root symlink, previously causing direnv to be constantly
reinvoked for some projects.

If merged, this should also fix issue #387.

---

Closes #387.

## Overview

I simply tried implementing the most obvious solution with guidance from #387. I have tested these by running the modified binary on my personal machine and with my personal projects.

Also, I am not sure if this should be classified as a feature or a fix.

## Checklist

### DIFF
--- a/src/build_loop.rs
+++ b/src/build_loop.rs
@@ -5,6 +5,7 @@ use crate::builder;
 use crate::daemon::LoopHandlerEvent;
 use crate::error::BuildError;
 use crate::nix::options::NixOptions;
+use crate::nix::StorePath;
 use crate::pathreduction::reduce_paths;
 use crate::project::roots;
 use crate::project::roots::Roots;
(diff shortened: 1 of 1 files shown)


leotaku:
Whoops, it seems like those CI errors are out of my control...

Issue #460, opened 2020-08-04
Does lorri work with nix flakes?


matthuszagh:
Based on the [instructions here](https://zimbatm.com/NixFlakes/#direnv-integration) and [here](https://github.com/direnv/direnv/wiki/Nix#nix-flakes), direnv is able to support flakes with a `use_flake()` function. However, when I tried this in a project with a `flake.nix` and `flake.lock` lorri complained about the lack of a `shell.nix` file. I haven't tried the `.envrc` suggestion in the second link because I'm not sure how this would work with lorri's requirement for `.envrc` to be `eval "$(lorri direnv)"`.

Does lorri support nix flakes? If so, how can it be used? If not, is support planned?

regnat:
I don't think lorri supports flakes atm, but you can use [flake-compat](https://github.com/edolstra/flake-compat/) to create a `shell.nix` that lorri can use, with

```nix
(import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
  src = ./.;
}).shellNix.default
```

Profpatsch:
Flakes are not a stable feature at the moment, they are not even mentioned in the nix manual.

We could start adding support for them, but afaik the format changes every few weeks at the moment, I don’t see the big selling point of interacting with experimental features.

regnat:
I think it would be a good time to start thinking about flakes for lorri − even without implementing it now − as the fact that it and the new cli are both experimental atm means that it's possible to upstream some lorri-specific logic into Nix itself (which will be much more painful once it gets stable, if only because lorri will have to be backwards compatible with the Nix-with-flakes-but-without-the-changes-that-make-it-easier-to-implement).

I'm not super-familiar with the internals of lorri, but my understanding of how lorri works is that it does essentially three things on the Nix side:

1. It builds an overriden version of the `shell.nix` where the builder is replaced by something that will just dump the environment
2. It adds a gc root for this overriden derivation (so that its inputs stays alive)
3. During the evaluation of this derivation, it hijacks a few builtins ((ab)using `scopedImport`) to make them trace their own execution (to know which files to watch)

On the Nix side,
1. Nix now has `nix print-dev-env` which seems to do the same thing
2. `nix print-dev-env` doesn't have a way to add a gc root, but it could be easily added (it's just a matter of plumbing as the code already has everything for that, the gc root is just discarded at the very last moment)
[...]

PR #506, opened 2021-07-25
Better error msg for flakes

gilescope:
While we don't have native support for flakes yet this PR suggests the workaround suggested in https://github.com/target/lorri/issues/460 if and only if there's no shell.nix and there's a flake.nix found.

I thought about creating the file automatically but I think just suggesting it is nicer.


gilescope:
wrong repo

